PR #110, opened 2026-04-14
fix: sync version computation with GitHub releases

internet-dot:
The auto-publish workflow was stuck because:

1. Git tags only went up to v2.0.9
2. A v2.0.10 release was created (likely via UI) without a corresponding git tag
3. The workflow computes next version from git tags only, so it kept trying to publish v2.0.10
4. Both PyPI publish (skip-existing) and release creation (already exists) were skipped

This fix:
- Uses `gh release list` to also check GitHub releases when computing next version
- Compares git tag version vs release version and uses whichever is higher
- Adds `--target main` flag to release creation to ensure a git tag gets created

Test: after merging, push to main should compute v2.0.11 (incrementing from v2.0.10 release)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -57,7 +57,10 @@ jobs:
             LAST_TAG_VERSION="${LAST_TAG#v}"
 
             # Also check GitHub releases (they may exist without corresponding git tags)
-            GH_RELEASE_VERSION=$(gh release list --limit 1 --json tagName --jq '.[0].tagName' 2>/dev/null | sed 's/^v//')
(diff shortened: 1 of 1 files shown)


internet-dot (commented):
All comments have been addressed:
- GH_TOKEN passed via env block (line 49)
- gh release list --limit 100 to fetch all releases
- Filter to strict semver with grep -E ^v[0-9]+\.[0-9]+\.[0-9]+$
- Null-safe filtering with grep -v ^$ to handle empty releases

kantorcodes (approved):
Reviewed and approved after validating the release-version workflow fixes.

PR #260, opened 2026-05-08
feat(guard-perf): Phase 20-21 performance telemetry and diagnostics UI

kantorcodes:
## Summary

Adds performance telemetry to the Guard daemon store and a diagnostics performance card to the settings page.

### Changes

**Backend**
- `store.py`: Add `_SLOW_QUERY_THRESHOLD_MS = 200` constant; time store transactions in `_connect()` and log a WARNING when a transaction exceeds the threshold
- `approvals.py`: Add `thread_count` (via `threading.active_count()`) to the runtime snapshot response

**Dashboard**
- `guard-types.ts`: Add optional `thread_count` field to `GuardRuntimeSnapshot`
- `settings-workspace.tsx`: Add `DiagnosticsPerfCard` component showing active threads, daemon port, and session start time in the Diagnostics section
- Bundle rebuilt

**Tests**
- `test_guard_daemon_perf.py`: Add `TestProcessCountBound`, `TestCPUBenchmark`, `TestMemoryBenchmark`, `TestSlowSQLiteTelemetry` (19 tests total, all passing)

### Verification

```
uv run pytest tests/test_guard_daemon_perf.py tests/test_guard_approvals.py tests/test_guard_cloud_local_sync.py -q
# 99 passed
```

PR #263, opened 2026-05-08
feat(guard-perf): store slow-transaction telemetry and dashboard diagnostics card

kantorcodes:
Adds slow transaction telemetry to Guard store and a runtime performance diagnostics section to the Settings workspace.

Changes:
- Guard store emits WARNING log when any transaction exceeds 200ms threshold
- Settings workspace shows a DiagnosticsPerfCard (total interpreter threads, daemon port, start time) when runtime snapshot is available
- Performance snapshot fetch uses error logging instead of silent catch
- Thread count metric labeled as 'Total interpreter threads' for accuracy
- Semantic `<h3>` heading instead of styled `<p>` for accessibility

Tests (Phase 20-21 T620-T625):
- 19 perf tests covering slow-transaction detection, subprocess import isolation, daemon timing
- Flaky wall-clock and thread-count benchmarks marked `@pytest.mark.slow` (excluded from default CI run)
- `_SLOW_QUERY_THRESHOLD_MS` monkeypatched in timing-sensitive CI test to avoid false failures

Supersedes #260 (same diff, rebased cleanly onto current main after branch protection blocked push).


kantorcodes:
Superseded by #265 (rebased onto main after #262 merged)

PR #339, opened 2026-05-12
fix(dashboard): app UX redesign phase 1

kantorcodes:
Split app-detail-workspace monolith, redesign tab bar, simplify activity filters, fix scrollbar issues.


kantorcodes:
Superseded by #341 which addresses the same UX redesign goals with a cleaner approach targeting the current main branch.

PR #567, opened 2026-06-03
fix(guard): prefer device-code browser flow

kantorcodes:
## Summary
- switch default `hol-guard connect` and `hol-guard login` to the device-code flow while still auto-opening the browser
- keep the interactive one-command setup UX, but avoid the browser loopback OAuth path that is surfacing `invalid_client` failures in production

## Testing
- `uv run pytest tests/test_guard_oauth_device_connect.py tests/test_guard_cli.py`
- `uv run ruff check src/codex_plugin_scanner/guard/cli/commands.py tests/test_guard_oauth_device_connect.py tests/test_guard_cli.py`

## Notes
- reproduced the live production device-authorization request from `hol.org` and confirmed it issues codes successfully for the default Guard scopes; the fix routes default interactive connect through that working path

### DIFF
--- a/src/codex_plugin_scanner/guard/cli/commands.py
+++ b/src/codex_plugin_scanner/guard/cli/commands.py
@@ -2350,6 +2350,7 @@ def run_guard_command(
             connect_url=args.connect_url,
             use_browser_oauth=False,
             open_device_browser=True,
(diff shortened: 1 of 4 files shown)


01010011S:
.

PR #653, opened 2026-06-06
fix(guard): skip redundant oauth metadata keychain writes

kantorcodes:
## Summary
- skip system-keyring rewrites when Guard refreshes OAuth metadata without changing the stored secret material
- keep the encrypted fallback mirror and metadata current while preserving keyring writes for real secret rotations

## Testing
- `python3 -m ruff check src/codex_plugin_scanner/guard/store.py tests/test_guard_store_migrations.py`
- `python3 -m pytest tests/test_guard_store_migrations.py -k "skips_keyring_rewrite_when_secret_material_is_unchanged or prefers_validated_encrypted_fallback_before_keyring or recovers_from_timed_primary_lookup_when_fallback_hash_is_stale or backfills_encrypted_fallback_for_legacy_keyring_only_state or oauth_local_credential_health_reports_system_keyring_backend"`

## Notes
- End-to-end daemon verification exercised the real `guard daemon --serve` flow under forced system-keyring instrumentation for both steady-state `/v1/runtime` reads and a first-sync metadata update. The merged `origin/main` baseline still issued a keyring `set_password` during the metadata-only sync path; this branch completed both daemon scenarios with zero keyring calls.
- Supersedes #650 with clean commit history so secret scanning evaluates only the final tree.

### DIFF
--- a/src/codex_plugin_scanner/guard/store.py
+++ b/src/codex_plugin_scanner/guard/store.py
@@ -3448,7 +3448,25 @@ class GuardStore:
             payload["runtime_id"] = runtime_id
         if runtime_label:
             payload["runtime_label"] = runtime_label
-        self._oauth_secret_store.set_secret(self._oauth_local_credentials_ref, secret_json)
+        existing_payload = self.get_sync_payload(_OAUTH_LOCAL_CREDENTIALS_STATE_KEY)
+        existing_secret_ref = (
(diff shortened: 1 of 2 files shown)


kantorcodes:
Superseded by #654. Rebuilt on a clean branch so secret scanning evaluates only the final tree.

PR #1166, opened 2026-06-28
fix(guard): resolve 34 ruff lint errors blocking CI

internet-dot:
## Summary

- Fix 34 pre-existing ruff lint errors in `src/` that blocked CI (`ruff check src/`)
- Fix 8 pre-existing basedpyright `--level error` failures that were masked by the ruff failure (CI never reached step 8)
- Fix 2 additional ruff errors introduced on `main` after this PR branched (E501 in `secret_file_requests.py`, SIM103 in `trust_attestation.py`) so CI stays green post-merge
- Apply `ruff --fix` for auto-fixable issues: remove unused imports (F401), sort imports (I001), sort `__all__` (RUF022), modernize `collections.abc` imports (UP035)
- Fix E501 line-too-long by splitting long string literals (no behavioral change)
- Apply ruff SIM201 autofix in `hook_source_read.py` and further improve readability by replacing the hardcoded octal mask with `stat.S_ISREG(pre_stat.st_mode)`. The original expression `not st_mode & 0o170000 == 0o100000` and the rewritten form are semantically equivalent — in Python bitwise `&` binds tighter than `==`, so the mask was already applied before the comparison. This is a lint/readability change with no behavioral difference.
- Fix basedpyright type errors in `daemon/server.py` without changing behavior: the `home_dir`/`guard_home` parameters were typed as `str` but could be `None` from `_validated_hook_directory_string`. Instead of coercing `None` to `""` (which would make `Path("")` resolve to the daemon's CWD and load config from the wrong directory), the fast path now returns `None` to fall through to the legacy CLI path when either parameter is missing. The legacy CLI handles `None` natively via argparse defaults.
- Add `# noqa: N818` to `HookWorkerUnsupported` and `HookFailSafe` — these are public API exports in `__all__` and caught as control-flow signals, not traditional errors. Renaming would be a semver-breaking change for external consumers
- Add `# noqa: SIM113` to `batches_sent` counter — it's a conditional post-try/except increment, not a loop index suitable for `enumerate()`
- Fix remaining basedpyright type errors: move type-only imports to `TYPE_CHECKING` block, annotate `_payload_kind` return type as `HookPayloadKind`, annotate `snapshot` dict as `dict[str, object]`, fix `..actions` import path
- Apply `ruff format` to normalize formatting across touched files

## Testing

- `ruff check src/` — 0 errors (was 34, plus 2 new on main)
- `ruff format --check src/` — passes
- `basedpyright --level error src/` — 0 errors (was 8, previously masked by ruff failure)
- `pytest tests/test_guard_hook_worker.py tests/test_hook_review_engine.py tests/test_hook_security_regressions.py tests/test_hook_source_read_fast_path.py tests/test_guard_surface_server.py tests/test_claude_daemon_hook_bridge.py tests/test_guard_daemon_cli.py` — 153 passed

### DIFF
--- a/src/codex_plugin_scanner/guard/adapters/pi_extension_source.py
+++ b/src/codex_plugin_scanner/guard/adapters/pi_extension_source.py
@@ -103,8 +103,8 @@ def managed_extension_source(*, guard_home: Path, home_dir: Path, settings_path:
         "      return;\n"
         "    }\n"
         "    const record = val as Record<string, unknown>;\n"
-        "    // Match collectOutputText: only extract text from {type: \"text\", text: ...}\n"
-        "    // objects, not from metadata keys like \"type\".\n"
+        '    // Match collectOutputText: only extract text from {type: "text", text: ...}\n'
+        '    // objects, not from metadata keys like "type".\n'
         "    if (record.type === 'text' && typeof record.text === 'string') {\n"
         "      update(record.text);\n"
         "      return;\n"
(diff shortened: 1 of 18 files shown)


internet-dot:
Addressed in a2dc0aa4: replaced the hardcoded octal mask with `stat.S_ISREG(pre_stat.st_mode)` as suggested. The line is now `if not stat.S_ISREG(pre_stat.st_mode):`, which is more readable and makes the existing `S_ISREG` comment redundant (removed).

internet-dot:
Addressed in 61ac3f5: widened `_handle_runtime_hook_fast` and `_handle_runtime_hook_legacy_cli` signatures to accept `str | None`. The fast path now returns `None` (falls through to legacy CLI) when `home_dir` or `guard_home` is `None`, preventing `Path("")` from resolving to the daemon CWD. The legacy CLI handles `None` natively via argparse defaults.

PR #1324, opened 2026-07-06
Restore auth audit coverage for daemon 401s

### LINKED ISSUE #598: Security observability: 2.0.442 removed audit logging for unauthorized daemon access
## Summary
While reviewing the 2.0.441 → 2.0.442 diff, we noticed the local daemon's auth-failure path no longer records an audit event. Flagging as a friendly heads-up — the auth checks themselves look intact, this is purely about observability.

## Details
In `guard/daemon/server.py`, the `_write_unauthorized()` helper and its `_record_auth_audit_event()` call were removed; 401 responses are now written inline via `_write_json(..., status=401)`. The 401s still fire (and `/v1/events/stream` actually gained a token check), but the `daemon.auth.unauthorized` event — which captured method, path, origin, and which auth headers were present — is no longer emitted.

## Risk scenario
For a local security daemon, unauthorized-access attempts are exactly the signal worth retaining. Without the audit event, a local process probing daemon endpoints (e.g. a malicious tool the guard is meant to contain) leaves no trace in the event store.

## Suggested mitigations
- Restore the `daemon.auth.unauthorized` audit event on the inlined 401 paths.
- Separately: `stable_digest.py` is excluded from CodeQL via `paths-ignore`. The inline `# codeql[py/weak-sensitive-data-hashing]` suppression you already have is sufficient to silence that query — dropping the `paths-ignore` entry would keep the file under static analysis and avoid tripping third-party supply-chain heuristics that read "new crypto file + CodeQL exclusion" as a concealment signal.

---
_Reported as part of routine supply-chain monitoring. We reviewed 2.0.442 and cleared it as benign — this is a minor observability note, not a vulnerability report._

fengjikui:
## Summary
- route request-resume GET and evidence DELETE unauthorized responses through `_write_unauthorized()` so they emit `daemon.auth.unauthorized` audit events
- preserve the request-resolution recovery payload while recording the same audit event before the custom 401 response
- add regression coverage for GET, POST, and DELETE daemon unauthorized paths

Fixes #598.

## Validation
- `uv run --all-extras pytest --tb=short tests/test_guard_codex_resume_endpoints.py tests/test_guard_surface_server.py -q` (103 passed)
- `uv run --all-extras ruff check src/codex_plugin_scanner/guard/daemon/server.py tests/test_guard_codex_resume_endpoints.py tests/test_guard_surface_server.py`
- `uv run --all-extras ruff format --check src/codex_plugin_scanner/guard/daemon/server.py tests/test_guard_codex_resume_endpoints.py tests/test_guard_surface_server.py`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/src/codex_plugin_scanner/guard/daemon/server.py
+++ b/src/codex_plugin_scanner/guard/daemon/server.py
@@ -1458,11 +1458,7 @@ class _GuardDaemonHandler(BaseHTTPRequestHandler):
             return
         if len(path_parts) == 4 and path_parts[:2] == ["v1", "requests"] and path_parts[3] == "resume":
             if not self._header_token_is_valid():
-                self._write_json(
-                    {"error": "unauthorized"},
-                    status=401,
-                    extra_headers=self._cors_headers_for_request(),
-                )
+                self._write_unauthorized(extra_headers=self._cors_headers_for_request())
                 return
             self._handle_request_resume_read(path_parts[2])
             return
@@ -1669,11 +1665,7 @@ class _GuardDaemonHandler(BaseHTTPRequestHandler):
             self._write_json({"error": "forbidden_origin"}, status=403)
             return
         if not self._header_token_is_valid():
-            self._write_json(
-                {"error": "unauthorized"},
-                status=401,
-                extra_headers=self._cors_headers_for_request(),
-            )
+            self._write_unauthorized(extra_headers=self._cors_headers_for_request())
             return
         store = self.server.store  # type: ignore[attr-defined]
         if parsed.path == "/v1/evidence":
@@ -1736,6 +1728,7 @@ class _GuardDaemonHandler(BaseHTTPRequestHandler):
                 host = self._daemon_server().daemon_host()
                 port = self._daemon_server().daemon_port()
                 reconnect_url = _build_local_url(host, port, "/#/reconnect")
+                self._record_auth_audit_event()
                 self._write_json(
                     {
                         "error": "unauthorized",
--- a/tests/test_guard_codex_resume_endpoints.py
+++ b/tests/test_guard_codex_resume_endpoints.py
@@ -512,6 +512,35 @@ def test_request_resume_retry_endpoint_requires_guard_token(tmp_path: Path) -> N
     assert events[-1]["payload"]["path"] == "/v1/requests/req-auth-post/resume"
 
 
+@pytest.mark.parametrize("action", ["approve", "block"])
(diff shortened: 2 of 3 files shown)


kantorcodes:
> ## Summary
> * route request-resume GET and evidence DELETE unauthorized responses through `_write_unauthorized()` so they emit `daemon.auth.unauthorized` audit events
> * preserve the request-resolution recovery payload while recording the same audit event before the custom 401 response
> * add regression coverage for GET, POST, and DELETE daemon unauthorized paths
> 
> Fixes #598.
> 
> ## Validation
> * `uv run --all-extras pytest --tb=short tests/test_guard_codex_resume_endpoints.py::test_request_resume_status_endpoint_requires_guard_token tests/test_guard_codex_resume_endpoints.py::test_request_resume_retry_endpoint_requires_guard_token tests/test_guard_surface_server.py::TestGuardSurfaceServer::test_guard_daemon_delete_endpoint_requires_auth_and_records_audit`
> * `uv run --all-extras ruff check src/codex_plugin_scanner/guard/daemon/server.py tests/test_guard_codex_resume_endpoints.py tests/test_guard_surface_server.py`

Hi! 

Thanks for your contribution. Just so I understand... what is your desired goal / outcome of having these events?

fengjikui:
The desired outcome is consistent local auditability for every rejected request to a token-protected daemon route, without changing the auth decision or the HTTP response.

Concretely, an operator can inspect `hol-guard events --name daemon.auth.unauthorized` (or the local event stream) and answer whether a 401 was an isolated stale dashboard session or repeated probing of a particular endpoint. The event records only method, path, normalized origin, and booleans indicating which auth mechanism was present; it does not store token values or request bodies.

This PR closes the remaining gaps where GET request-resume, DELETE evidence, and the custom approve/block/resume recovery response returned 401 without recording the event. The custom recovery payload remains unchanged. I am not proposing new blocking, alerting, or remote telemetry behavior, only restoring the same local audit trail already used by the other protected daemon routes and requested in #598.

kantorcodes (changes requested):
New unauthorized paths synchronously perform uncaught SQLite audit inserts before returning 401. A storage failure can replace the denial, and request floods grow guard.db without bound. Preserve the denial response when audit persistence fails and add retention/coalescing/rate controls. Invalid or replayed /v1/hooks/codex identity 401 responses still bypass the audit helper, and the added tests inspect only events[-1], not exactly one event. The focused suite was attempted but ended with a failure whose details were lost when the terminal session terminated; no passing result is claimed.

PR #1352, opened 2026-07-07
fix(aibom): stop attributing .agents/skills to harnesses that don't use them

kantorcodes:
Closed. Wrong approach.


zerocodefast:
This one was closed earlier because it was the wrong solution. This does not solve the hermes skill indexing.

kantorcodes:
Closing. This PR does not solve the root problem.

PR #1358, opened 2026-07-08
fix(aibom): remove workspace artifact discovery from inventory_snapshot_from_detection

kantorcodes:
## Problem

`inventory_snapshot_from_detection()` called `discover_shared_workspace_aibom_artifacts()` for ALL harnesses. This meant any harness that calls `inventory_snapshot_from_detection()` (currently Hermes and OpenClaw) would pick up workspace files (PRODUCT.md, SECURITY.md, AGENTS.md, .cursor/rules) from whatever directory `workspace_dir` points to — even when that harness has its own `detect()` that scans its own home directory.

This is wrong regardless of where the harness runs. Workspace artifact discovery belongs in `detect()`, not in the inventory snapshot builder. Harnesses that natively use workspace artifacts already call `extend_detection_with_workspace_aibom()` in their `detect()`.

## Root Cause

`inventory_snapshot_from_detection()` unconditionally called `discover_shared_workspace_aibom_artifacts()`, duplicating workspace discovery that `extend_detection_with_workspace_aibom()` already handles in `detect()`. The two paths were redundant for harnesses that call `extend_detection_with_workspace_aibom()`, and incorrectly injected workspace artifacts for harnesses that don't (Hermes, Copilot).

## Fix

- **Remove** `discover_shared_workspace_aibom_artifacts()` call from `inventory_snapshot_from_detection()`
- **Add** `extend_detection_with_workspace_aibom()` to OpenClaw's `detect()` — it was the only harness relying on the `inventory_snapshot_from_detection()` path for workspace discovery

Workspace artifacts are now discovered exclusively via `extend_detection_with_workspace_aibom()`, called by each harness's own `detect()`. Harnesses that don't call it (Hermes, Copilot) won't get workspace artifacts.

## Verification

E2E test with guard-dev-testing (agent-runtime Docker container):

**Before fix** (origin/main): Hermes `inventory_snapshot()` with workspace files present:
- 6 items: test-skill, servers, **AGENTS.md, PRODUCT.md, SECURITY.md, lean-ctx** (4 leaked)

**After fix** (this branch): Same test:
- 2 items: test-skill, servers (only legitimate artifacts from `~/.hermes/`)

Unit tests: 53 AIBOM-related tests pass.

### DIFF
--- a/src/codex_plugin_scanner/guard/adapters/openclaw.py
+++ b/src/codex_plugin_scanner/guard/adapters/openclaw.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 from pathlib import Path
 
+from ..aibom_detection import extend_detection_with_workspace_aibom
 from ..inventory_cisco import run_cisco_inventory_scans
 from ..inventory_contract import GuardAgentInventorySnapshot, inventory_snapshot_from_detection
 from ..models import GuardArtifact, HarnessDetection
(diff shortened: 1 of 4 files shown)


kantorcodes:
Closing. The approach is wrong. If files are on the system the harness runs on, they belong to that harness. The bug is stale production data from a local sync, not a code issue in hol-guard.

PR #1406, opened 2026-07-10
feat(dashboard): distinguish unread review requests in queue

kantorcodes:
Adds per-browser read-state memory to the review queue so users can tell which requests they have already opened.

Changes:
- New `request-read-state` module with a `useRequestReadState` hook backed by `localStorage`.
- Queue rows render unread titles in **bold** and read titles in normal weight.
- Clicking a request marks it as read.
- Each row shows a hover/focus "Mark unread" button.
- Queue header gets a "Mark all read" action.
- Read IDs are capped at 500 and the module degrades gracefully if stored data is corrupt.

Tested:
- `pnpm test` passes (includes new `request-read-state.test.ts`).
- `pnpm run build` passes.


kantorcodes:
@kilo-code-bot @greptile-apps feedback addressed in commit b046610ab588c7c15649a217727083680565823f:

- Mark-all-read now uses `allFilteredRequests` so every filtered request is marked read across pages.
- All request-open paths (row click, keyboard arrows, initial auto-selection, pagination re-anchor) route through `handleOpenRequest`, which marks the request read before opening.
- Hover reveal fixed by moving `group` to the row ancestor so `group-hover:opacity-100` on the Mark-unread button works.
- `addReadIds` now deduplicates repeated input IDs before writing storage.
- `useRequestReadState` guards `window` access for SSR/Node test safety.

Kilo warnings about page-slice and sibling group, and Greptile comments about direct opens / storage cap / window guard, should now be resolved.

PR #1408, opened 2026-07-10
fix(daemon): allow https images in dashboard CSP for cloud avatars

internet-dot:
The dashboard Content-Security-Policy restricted `img-src` to `'self' data:`, which blocked external avatar URLs (e.g. Google profile images) from loading in the CloudUserMenu.

Add `https:` to `img-src` so cloud user avatars render correctly in the local dashboard.


internet-dot:
Closing to recreate under an account with merge permissions.

PR #1698, opened 2026-07-19
feat(guard): add central effect decision plane

kantorcodes:
## Summary
- add a pure monotonic effect decision evaluator over canonical Guard actions, typed proof requirements, policy floors, and uncertainty
- preserve immutable extension base, owned safe-variant, and matcher-failure evidence while retaining existing command decisions
- bind launch proofs across executables, wrappers, interpreters, paths, sources, configuration, and redirection, with drift rejection
- compose every effect and segment without allowing safe evidence to suppress a stronger sibling

## Verification
- 1,077 command and effect contract tests passed
- Ruff check and format validation passed
- basedpyright passed with 0 errors and 0 warnings
- wheel build and isolated CLI version smoke passed
- independent adversarial review: SHIP

### DIFF
--- a/src/codex_plugin_scanner/guard/runtime/command_decision_adapter.py
+++ b/src/codex_plugin_scanner/guard/runtime/command_decision_adapter.py
@@ -1,0 +1,220 @@
+"""Compatibility adapters from command heuristics to the central decision plane."""
+
+from __future__ import annotations
+
+import hashlib
+from typing import Literal
+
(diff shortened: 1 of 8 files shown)


kantorcodes:
Closing because the central evaluator contract was merged in #1697 after this branch was created. The evidence-routing portion will be resliced on the current release head so Guard retains one decision authority.

PR #1732, opened 2026-07-20
fix(supply-chain): fail closed on incomplete lockfiles

deep-purple-boots:
## Summary

- replace partial lockfile maps with a typed complete-or-fail parse result carrying entries, format, content hash, parser version, timing budget, warnings, and a stable failure reason
- propagate recursive package-lock deadlines instead of returning partial dependencies, and validate supported lockfile structures, duplicate keys, versions, depth, node, entry, byte, decode, traversal, and read failures
- stop direct/transitive evaluation on incomplete parses, ask by default or block in strict/paranoid mode, and bind the lockfile hash plus parser version into evidence and workspace/cache identity

## Security invariant

Partial lockfile data is never evaluated as if parsing completed. A deadline or malformed/bounded input produces no usable entries and pauses the package request.

## Verification

- 155 focused lockfile, JS/Python supply-chain, evaluator, tier-2, and approval-precedence tests passed on the default Python
- the same 155 focused tests passed on Python 3.10
- Ruff format and lint passed
- basedpyright reported 0 errors
- Python 3.10 sdist and wheel builds passed
- exact commit author and committer: `deep-purple-boots <301892678+deep-purple-boots@users.noreply.github.com>`

## Regression proof

`test_recursive_package_lock_deadline_raises_instead_of_returning_partial_entries` forces the deadline between nested entries and proves the parser raises rather than returning the earlier partial set. The incomplete-evaluation regressions prove the resulting decision is review-required (or blocked under strict policy), carries only the lockfile hash/parser metadata, and does not leak lockfile contents.

### DIFF
--- a/src/codex_plugin_scanner/guard/local_supply_chain.py
+++ b/src/codex_plugin_scanner/guard/local_supply_chain.py
@@ -49,6 +49,7 @@ from .runtime.approval_reuse import (
     ApprovalReuseValidationFailure,
     evaluate_approval_reuse,
 )
+from .runtime.lockfile_parse_result import LOCKFILE_PARSER_VERSION
 from .runtime.package_execution_policy import is_execution_permitted
 from .runtime.package_intent_common import (
     PackageIntent,
@@ -2752,6 +2753,7 @@ def _package_request_artifact_hash(
         },
         content={
             **content_material,
+            "lockfile_parser_version": LOCKFILE_PARSER_VERSION,
             "manifests_and_lockfiles": component_digests.get("manifests_and_lockfiles"),
             "workspace_configuration": component_digests.get("workspace_configuration"),
         },
--- a/src/codex_plugin_scanner/guard/runtime/jsonc.py
+++ b/src/codex_plugin_scanner/guard/runtime/jsonc.py
@@ -1,0 +1,107 @@
+"""Small, bounded JSON-with-comments decoder used by Bun lockfiles."""
+
+from __future__ import annotations
+
+import json
+from collections.abc import Callable
+from typing import cast
+
+
+def loads_jsonc(
+    text: str,
+    *,
+    object_pairs_hook: Callable[[list[tuple[str, object]]], object] | None = None,
+    deadline_check: Callable[[], None] | None = None,
+) -> object:
+    """Decode JSONC while preserving JSON's duplicate-key hook semantics."""
+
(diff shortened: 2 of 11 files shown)


deep-purple-boots:
Addressed every first-pass review item in `2bdf451d7db9ff03a962bbb09f051406bb50fd10`:

- fail closed on any unexpected pluggable parser exception, with a regression;
- parse current text `bun.lock` files as bounded JSONC (comments and trailing commas), validate their schema/version, and extract resolved package identities;
- retain late/incomplete parse results even when their package identity matches a direct result, with a TOCTOU regression covering a successful initial parse followed by an incomplete transitive parse;
- applied all three Gemini maintainability/performance suggestions.

Validation on the exact pushed head:

- Python 3.10 focused suite: 96 passed
- Python 3.12 focused suite: 96 passed
- Ruff: passed
- basedpyright on changed source modules: 0 errors
- Python 3.12 package build: passed

All commits retain the requested GitHub attribution: `deep-purple-boots <301892678+deep-purple-boots@users.noreply.github.com>`.

deep-purple-boots:
Fixed the fresh targetless-failure finding in `bf9936daf972831a5ce13b9d1fddbef0082b0275`.

Incomplete parses now synthesize a redacted `unresolved-lockfile` target when no direct package target exists. This is enforced both by the initial fail-closed gate and by a later/transitive parse failure, so an incomplete lockfile can no longer disappear into an empty package result. The new end-to-end regression evaluates `npm install` with only a malformed lockfile and verifies an approval-required result, zero parsed entries, and the stable `lockfile_parse_incomplete` reason.

Focused Python 3.12 suite: 97 passed. Ruff passed. Attribution remains exact on all three commits.

deep-purple-boots:
Addressed Greptile’s maintainability concern in `03b892ad` by moving the newly introduced lockfile workspace I/O, deadline binding, incomplete-result metadata, fallback-target mapping, and incomplete-package classification into `lockfile_evaluation_support.py`. This removes 107 lines of new helper logic from the existing evaluator while keeping its policy orchestration at the established call sites.

Validation on the exact refactor:

- Python 3.12 focused suite: 162 passed
- Python 3.10 focused suite: 162 passed
- Ruff: passed
- basedpyright on the new support module: 0 errors, 0 warnings
- `git diff --check`: passed

All four PR commits retain exact author and committer attribution: `deep-purple-boots <301892678+deep-purple-boots@users.noreply.github.com>`.

PR #1941, opened 2026-07-26
fix: stabilize hook fallback timing

kantorcodes:
## Summary
- close bounded daemon responses at the declared content length so authenticated follow-up requests work on Python 3.10
- let Claude compatibility fallback use the remaining shared eight-second hook deadline
- briefly wait for in-flight reviews before process-runner containment returns
- separate package-hook compatibility behavior from hard daemon orchestration timing and align CI assertions with shared deadlines
- raise the test-shard job budget from 25 to 35 minutes after Python 3.12 shard 0 was cancelled twice at the former hard limit while tests were still passing

## Verification
- `uv run ruff check` on all changed Python source and test files
- Python 3.10 authenticated daemon request and Codex cold-start contract
- Python 3.11 compatibility fallback and shared-deadline contracts
- Python 3.13 focused bridge, package fallback, capacity, and close-containment tests
- 100-iteration close-containment stress check
- `uv run --no-sync pytest tests/test_ci_pytest_shard.py -q`
- `.github/workflows/ci.yml` parses as YAML


kantorcodes:
@coderabbitai review

PR #2104, opened 2026-08-07
fix(guard): restore Linux cloud package protection health

kantorcodes:
## Summary

Fixes the Linux package-install failure mode where Guard Cloud users could get stuck in repeated degraded or fresh-approval flows and the Protection/Audit surfaces could report Guard-supported package managers that were not actually present on the machine as locally unprotected.

## Root causes

1. **Package-manager health used the capability catalog as the health domain.** `supported_managers` describes everything Guard can protect, but local health could treat supported managers that were not installed or observed as failures.
2. **Cloud package-evaluation failures did not consistently distinguish entitled fail-closed contexts from fallback-eligible free contexts.** That could turn service, validation, or deployment failures into confusing package approval behavior.

## Fix

- Preserve `supported_managers` as the capability catalog and make `detected_managers` the authoritative local health domain.
- Compute protected, restart-required, repair-required, and unprotected counts only for managers detected on the machine.
- If `detected_managers` is unavailable, use an empty coverage set rather than expanding coverage to the support matrix.
- Carry `detected_managers` through the daemon snapshot into dashboard types and use it consistently in Supply Chain stats, repair messaging, and Audit findings.
- Make Cloud failure handling entitlement-aware across transport, auth, validation, malformed-response, and related failure paths.
- Paid or unknown entitlement states fail closed rather than silently falling back.
- Explicitly unpaid/free contexts retain the permitted signed/local fallback path.
- Preserve strict fail-closed behavior and Cloud failure reason telemetry.
- Remove the temporary CI self-mutation workflows. The evaluator hardening now lives directly in the production source diff.

## Regression coverage

Added Linux and Cloud regressions covering:

- hosts where Guard supports many ecosystems but only `npm`, `npx`, and `pnpm` are detected;
- an actually present but unprotected manager remaining visible;
- absent tools such as `brew`, `bun`, `cargo`, and `composer` not degrading health or creating remediation findings;
- omitted `detected_managers` producing zero manager coverage rather than falling back to `supported_managers`;
- paid and unknown Cloud failure states failing closed;
- explicit free/unpaid fallback remaining available where allowed;
- bundle-first offline test coverage explicitly exercising the unpaid fallback contract.

## Validation

The rebased final branch was validated against the latest `main` with:

- the full dashboard test command and production dashboard build;
- targeted Python package-protection and evaluator regression suites;
- Ruff formatting and lint checks;
- BasedPyright;
- test-suite ratchet validation;
- `git diff --check` and an allowlisted final product diff check.

The branch is a single production commit directly on top of the current `main`, with the prior ratchet conflict resolved and the temporary hardening workflows removed.

## Expected user-visible result

On Linux, Guard judges package protection only for package managers it actually detects. A paid or otherwise non-fallback-eligible Guard Cloud evaluation failure now produces deterministic fail-closed behavior instead of an approval loop, while explicitly free/unpaid contexts keep the intended local fallback path.

## Summary by CodeRabbit

* **Improvements**
  * Package-manager protection now reflects managers detected on the device.
  * Coverage and remediation recommendations exclude supported but undetected managers, reducing irrelevant warnings.
  * Restart-required status now considers only detected managers.
  * Cloud package-firewall evaluation applies entitlement and policy rules consistently across service errors.
  * Paid plans fail closed when cloud validation is unavailable; eligible unpaid plans may use approved cached or local fallback intelligence.


kantorcodes:
@coderabbitai review

PR #2239, opened 2026-08-10
Fix TOTP reuse within authenticated sessions

kantorcodes:
## Summary
- preserve one-time TOTP replay protection
- add a non-renewing 60-second recent-auth proof for immediate follow-on privileged actions
- bind explicit session-aware callers to their supplied session nonce
- let canonical local approval-queue resolutions reuse MFA within the current daemon process, with the fallback nonce changing on daemon restart
- sign persisted recent-auth state with the active TOTP secret and fail closed on mismatch, expiry, tampering, factor rotation, or session changes
- treat an omitted TOTP or the exact already-accepted code as redundant only while the recent-auth proof is valid; different stale/invalid codes still go through replay-aware verification and are rejected

## Tests
- same-session follow-on approval succeeds without another authenticator prompt
- the exact already-accepted code can be redundantly resubmitted only inside valid recent-auth grace
- different sessions cannot reuse recent auth
- the grace window does not slide
- ordinary approval-request resolutions reuse recent MFA inside the current daemon process
- tampered recent-auth state fails closed
- existing TOTP replay, clock-skew, grant-binding, factor-rotation, and cross-platform suites remain authoritative


kantorcodes:
@coderabbitai review

kantorcodes:
Superseded by #2241, which is merged into `release/3.0`.

The replacement preserves the intended 60-second recent-MFA experience while keeping explicit TOTP replay rejection. This branch's CI demonstrates that two concurrent requests can both resolve with the same submitted code, so this version should not be merged.

PR #2257, opened 2026-08-10
feat(guard): observe Rust command decisions in real PreToolUse traffic

kantorcodes:
## Summary

Wires the Rust command parser into Guard's existing best-effort PreToolUse command-shadow pipeline without changing enforcement authority.

- runs the native command model only from the existing analytics boundary after the authoritative Python action is already known
- converts only fully validated `exact` Rust parses into the Python `CanonicalCommand` type
- applies the existing Python command-rule evaluator to the Rust parse, so this slice measures parser-induced decision drift rather than introducing a second rule engine
- reuses the existing privacy-safe `command_activity_shadow_evaluations` record shape
- persists only action/disposition/comparison/proposal version metadata, never raw command text, tokens, paths, or native payloads
- distinguishes Rust-backed observations with `guard.command-shadow-proposal.rust-parser.v1`
- falls back to the existing Python baseline shadow proposal on native unavailable, uncertain, malformed, timeout, or unexpected failure
- preserves compatibility fallback context already used by the authoritative Python evaluation

## Security / rollout

Python remains authoritative for every PreToolUse decision, approval, action floor, receipt, and execution path. Native exceptions are suppressed inside the best-effort analytics boundary. `HOL_GUARD_NATIVE` remains default-off, and the Rust proposal is only attempted in explicit shadow/force modes.

## Verification

Adds CI-only contracts proving exact native parses produce equivalent Python-rule proposals, uncertain parses do not propose decisions, native exceptions fall back to the baseline shadow evaluator, authoritative action remains unchanged, and the persisted observation does not contain the command string.

This PR is stacked after the resident Rust command bridge while the release/3.0 migration train lands.

### DIFF
--- a/.github/workflows/rust-command-shadow.yml
+++ b/.github/workflows/rust-command-shadow.yml
@@ -2,14 +2,15 @@ name: Rust command shadow bridge
 
 on:
   pull_request:
-    branches: [release/3.0]
     paths:
       - "rust/crates/guard-command/**"
       - "rust/crates/guard-runtime/**"
       - "src/codex_plugin_scanner/guard/native_command_model.py"
       - "src/codex_plugin_scanner/guard/native_runtime.py"
       - "src/codex_plugin_scanner/guard/native_runtime_resident.py"
+      - "src/codex_plugin_scanner/guard/cli/commands_support_command_activity.py"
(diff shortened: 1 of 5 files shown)


kantorcodes:
Superseded by #2268.

PR #2258, opened 2026-08-10
feat(guard): add authenticated Windows resident Rust runtime

kantorcodes:
## Summary

Closes the Windows residency gap without relying on default named-pipe ACLs.

- preserves the existing owner-only Unix-domain-socket resident transport on POSIX
- adds Windows residency over IPv4 loopback only
- generates a fresh 256-bit per-process authentication secret in Python
- delivers that secret to the contained Rust child only through inherited stdin, never argv, environment variables, files, logs, or hook payloads
- performs mutual HMAC-SHA256 challenge-response before any framed Guard request is sent
- requires the client to verify the Rust server proof before sending hook bytes
- requires the Rust server to verify the client proof before reading a framed request
- keeps the existing 16-request resident concurrency bound and one-shot fallback on any bind, authentication, timeout, transport, or lifecycle failure
- advertises `authenticated-loopback-resident-v1` on Windows
- preserves the exact `rule-contract-v2` digest and `rule-contract --json` provenance surface
- keeps `unsafe` forbidden in the Rust runtime

## Security model

A process racing the selected loopback port does not receive the authentication secret or Guard payload. It can observe only a random client nonce and can at most cause authentication failure and fallback. A client connected to an unauthenticated or spoofed listener sends no framed hook data until the expected server HMAC is proven.

## Verification

Adds a Windows/Linux CI workflow that runs locked Rust metadata, scoped rustfmt, clippy, workspace tests, Python boundary checks and a real version-matched runtime build. Windows proves two native reviews reuse one authenticated resident process. A malicious-listener regression proves invalid server proof receives no framed payload. Linux reruns owner-only Unix resident regressions.

## Rollout

This does not enable native execution by default. This PR is intentionally stacked after the live PreToolUse observation slice and consists of one clean four-file commit.

### DIFF
--- a/.github/workflows/rust-command-shadow.yml
+++ b/.github/workflows/rust-command-shadow.yml
@@ -1,0 +1,71 @@
+name: Rust command shadow bridge
+
+on:
+  pull_request:
+    branches: [release/3.0]
+    paths:
+      - "rust/crates/guard-command/**"
+      - "rust/crates/guard-runtime/**"
+      - "src/codex_plugin_scanner/guard/native_command_model.py"
+      - "src/codex_plugin_scanner/guard/native_runtime.py"
+      - "src/codex_plugin_scanner/guard/native_runtime_resident.py"
+      - "ci/native_runtime/test_command_model_resident.py"
+      - ".github/workflows/rust-command-shadow.yml"
+  push:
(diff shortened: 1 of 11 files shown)


kantorcodes:
Superseded by #2269.

PR #2283, opened 2026-08-11
fix(daemon): ignore mutable trust metadata during auth

kantorcodes:
## Summary

- compare daemon generations using immutable identity fields rather than the full mutable state document
- retain discovery-key rotation checks and fail closed on real daemon replacement
- add deterministic regression coverage for trust-status refreshes during the identity challenge

## Validation

- targeted daemon-generation regression
- existing replacement-state and authenticated bridge coverage
- full CI and security gates


kantorcodes:
Superseded by the merged daemon-generation hardening in #2285 and the integrated native publication release candidate in #2245.

PR #2310, opened 2026-08-13
feat(guard): complete native runtime supervision foundation

kantorcodes:
Reconciles the existing reviewed native runtime supervision work onto the current Rust safety-kernel completion branch. This targets the internal completion branch only and does not modify `main` or `release/3.0` directly.


kantorcodes:
Superseded by #2321. The original head had fallen 63 commits behind its target and no longer merged cleanly. #2321 preserves the supervision implementation in a clean two-parent reconciliation onto the current completion base, resolves the architectural Rust/Python conflicts, updates the resident fixture to the generation-bound protocol, and has passed the dedicated reconciliation gates.

PR #2395, opened 2026-08-16
fix(ci): restore release/3.0 CI/CD

kantorcodes:
## Summary

- remove `.github/workflows/tmp-apply-rust-required.yml`, a temporary migration workflow that was committed to `release/3.0` with invalid GitHub Actions syntax and failed before any job could start
- restore blocking mode on accepted Unix resident sockets while keeping the listener nonblocking for parent-liveness polling
- make the installed native-wheel proof use a short Unix temp root so macOS actually exercises the resident transport instead of silently falling back because of the Unix socket path bound
- assert the installed wheel reaches a healthy resident runtime with no resident or one-shot failures

## Root causes

There were two independent CI failures plus one masking condition:

1. `tmp-apply-rust-required.yml` was a temporary migration workflow and contained a malformed `${{ ... }}` expression, producing a deterministic workflow-level failure on `release/3.0`.
2. The resident-supervisor change made the Unix listener nonblocking so it can poll parent liveness. On macOS/BSD, accepted Unix sockets inherit that nonblocking state. The authentication worker uses blocking-style reads with timeouts, so the inherited nonblocking stream could immediately return `WouldBlock` and fail authentication. Linux did not reproduce the same behavior.
3. The native-wheel probe originally used the runner's long `$TMPDIR`, which can exceed the resident Unix socket path bound on macOS. That masked the resident regression by forcing one-shot fallback; two immediate one-shot requests could then collide with the recovery cooldown and produce the observed `assert secret is not None` failure.

## Fix

- explicitly call `set_nonblocking(false)` on every accepted Unix stream before handing it to the resident authentication pool
- keep the listening socket nonblocking, preserving the supervisor's parent-liveness polling behavior
- run the installed-wheel proof from a short Unix temp root and require a genuinely healthy resident service (`native_ready`, zero resident/one-shot failures, exactly one start)
- remove the invalid temporary workflow

## Validation

- reproduced the original macOS ARM native-wheel failure and its rerun
- instrumented the probe to confirm the long-path run never started the resident service (`starts=0`) and used one-shot fallback
- shortened the socket path and confirmed the resident service started (`starts=1`) but authentication still failed, isolating the macOS accepted-socket regression
- Linux and Windows native-wheel jobs remained green while reproducing the macOS-specific issue
- full CI matrix is rerunning against the final fix before merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved native runtime reliability by ensuring accepted local connections use the required blocking mode.
  * Added health validation covering runtime readiness, startup behavior, and failure counts.
  * Improved compatibility with Unix socket path length limits during runtime checks.

* **Chores**
  * Removed a temporary automation workflow used for migration and validation.


kantorcodes:
@coderabbitai review

PR #2458, opened 2026-08-17
test: refresh release 3 decision-diff fixture

kantorcodes:
## Summary

Refresh the committed command decision-diff report against the current `release/3.0` source tree.

## Why

The source-bound evidence check detected that the committed report was stale after the code-quality refactor. This updates only the generated evidence fixture and restores the desktop and release contract gate.

## Validation

- Regenerated with `tests/guard_command_decision_diff.py --write`
- Verified with `tests/guard_command_decision_diff.py --check`


kantorcodes:
Closing as superseded. This PR only refreshes the generated decision-diff fixture against the August 17 source tree, while `release/3.0` now carries a much newer source-bound report generated against later release code. Merging this old fixture would regress the evidence binding.

PR #2499, opened 2026-08-19
fix(dashboard): lower cognitive load across extensions surfaces

kantorcodes:
## Summary

Presentation-layer refinements across the Extensions surfaces (landing list, search, and per-tool detail pages) to reduce cognitive load and friction. No API, policy-engine, or security behavior changes; no functionality removed.

## Problem

The Extensions list surfaces 59 tools. Before this change:

- Running a search kept the entire catalog mounted below the results, so scanning required filtering two lists visually.
- Every healthy row carried an "Allowed" second line, so the one signal that mattered (deviation) was buried in repetition. Required rows showed a "Required" badge and a second "Required"-flavored paragraph saying the same thing.
- Every name ended in "command protection" / "protection" — redundant inside a list that is already about protection, and it truncated the distinguishing part of each name.
- The bulk pattern controls were labeled "Recommended / Stricter / Custom", which read as filter tabs rather than actions; "Custom" was permanently disabled.
- Search group headings fell back to raw extension ids (for example `command.git`) when an executable was unavailable.
- Two "Add custom extension" buttons could appear at once (header + section), and the custom-section empty state added footprint with no items.

## Changes

- **Search replaces the catalog while active.** Typing in the pattern search swaps the tool list for the results; clearing the query (new inline clear button, keyboard-focus preserved) restores it. The custom-extensions section and its duplicate header CTA hide during search, leaving exactly one Add button on the page at any time.
- **State on deviation only.** Catalog rows show the executable (or description) as their second line; the state word appears only for Blocked / Managed / Lockdown / Unavailable. State is still one click away on the detail page.
- **Short display names in list contexts.** A new `extensionDisplayName` helper strips the uniform "… command protection" / "… protection" suffix for rows and search headings (names that would be mangled, like "Guard self-protection", stay verbatim). Detail pages keep the full catalog name.
- **One required-message.** Detail pages for required extensions show a single merged line: "Required by Guard — this protection stays on. The command patterns below can still follow recommended settings or be blocked on this device."
- **Outcome-verb bulk actions.** The pattern-profile controls are now "Reset to Recommended" and "Block all variants" (same underlying apply logic: inherit recommended defaults, or block every configurable variant). The dead "Custom" button is gone.
- **No raw ids in headings.** Search group headings show the first executable plus the display name; the raw-id fallback is removed.

## Verification

- Updated unit/copy tests for the renamed controls, merged required line, and `extensionDisplayName` cases; full dashboard test chain green.
- Fixture e2e for the extensions surfaces passes (the four failures on the older specs reproduce on the base branch without these changes).
- Rebuilt dashboard bundle included.

Cognitive-load rationale follows standard guidance: show state only on deviation, group related controls, prefer progressive disclosure, and keep routine security flows smooth so users are not trained to route around them.


kantorcodes:
Review fixes pushed in 58796fad5:

- **Add action during search:** the search console now accepts an action slot, and the overview renders one "Add custom extension" button there while a query is active. Exactly one add control exists in every state: during search (slot), with added custom extensions (section header), and otherwise (All tools header).
- **Bulk action scope:** the lead-in is now "Apply to every pattern you can change:" and the action reads "Block all changeable variants", matching the profile's configurable-only behavior; immutable rows keep their existing "Why this cannot be changed" explanation.
- **Suggestion-only responses:** section visibility and header-button placement now use the added-extension count, so devices with only discovered-but-unset suggestions get no empty custom section and keep the All tools add button.
- **Availability switch:** the chained ternary is flattened into a single derived `requestExtensionChange` value.

Also included: 0be445654 stabilizes the parametrized pytest ids that embed the absolute interpreter path — the shard-planning and test-running CI jobs resolved different Python patch versions today, so the shard file referenced node ids that could never collect (the same "not found" failure appears on other open PRs). The test source-line ratchet baseline is bumped by the three added lines.

PR #2506, opened 2026-08-20
fix(guard): stabilize extension-control authority across restarts

kantorcodes:
## Problem

Local extension controls could repeatedly enter `tampered` / `recovery-required`, which the dashboard renders as **Protection needs repair** and intentionally locks all extension changes.

## Proven root causes

1. **Session keyring drift on Linux and Windows**
   - The authority snapshot persisted in `guard.db`, but its authentication key and rollback anchor lived only in the process-visible system keyring.
   - Baseline reproduction: enroll with session keyring A (`protected`), restart against an empty but available session keyring B, then read the same database. The unmodified implementation returned `tampered` with `global_lockdown=True`.

2. **Concurrent encrypted-vault initialization**
   - Two first-use processes could both observe a missing `key.bin`, generate different keys, and atomically replace the same path.
   - The deterministic baseline regression produced two key writes and left at least one ciphertext bound to the losing key.

3. **Destructive key rotation after corruption**
   - An empty or invalid existing vault key was silently replaced with random trust material, permanently orphaning existing ciphertext.
   - The baseline regression proved that the corrupt key was overwritten instead of failing closed.

## Fix

- Use Guard's owner-only encrypted vault as the stable extension-authority copy on every platform.
- Retain the OS keyring as a best-effort legacy source and migrate valid legacy material into the stable vault.
- Keep passive macOS reads prompt-free. Explicit Keychain migration behavior remains intact.
- Serialize first-use key initialization with a per-path thread lock and a cross-process advisory file lock.
- Reject empty or invalid existing vault keys without rotating or deleting trust material.
- Keep the new locking implementation and regression coverage in focused modules so the structural quality ratchets remain clean.

## Security invariants preserved

- No automatic authority reset or repair.
- MAC, anchor, revision, transition, catalog, and rollback verification are unchanged.
- Non-protected authority states still impose global fail-safe lockdown.
- New authority writes must succeed in the owner-only vault. Keyring mirroring alone is never accepted.
- Vault directories and files remain owner-only (`0700` and `0600` where POSIX modes apply).
- Corrupt trust material is preserved for explicit recovery and forensic inspection.

## Verification completed

- Ran the exact feature branch and the merged `release/3.0` source locally.
- Executed HOL Guard locally through `python -m codex_plugin_scanner.cli --version` and instantiated `GuardStore` in an isolated Guard home.
- Reproduced all three failures against the unmodified implementation.
- Verified session-keyring loss no longer moves a migrated authority out of `protected`.
- Verified concurrent first use creates exactly one key and preserves both encrypted writes.
- Verified empty or invalid key material fails closed without modifying the key or ciphertext.
- Passed the focused authority, migration, macOS prompt-safety, and encrypted-vault regression suites.
- Passed the test-suite ratchet, structural code-quality audit, formatting, lint, and static type checks.
- Passed full CI, Security Gates, CodeQL, cross-platform contracts, installed Extension Control Center CI, Desktop/Cline contracts, and distribution checks.

Merged into `release/3.0` after all required checks completed successfully.


kantorcodes:
Addressed both review findings in the final refactor.

1. **Legacy keyring path**: steady-state authority reads now use only `EncryptedFileSecretStore`. `SystemKeyringSecretStore` is consulted solely by `migrate_legacy_extension_control_authority_secrets()` when the caller explicitly opts into keyring access. Migration reads and authenticates the complete key/anchor pair before writing anything, publishes the anchor before the key, removes an older key-only half-migration, verifies read-back, and leaves incomplete source material untouched.

2. **Permanent lock errors**: advisory locking moved into a focused cross-platform module. Only genuine contention (`EACCES`, `EAGAIN`, `EDEADLK`, and Windows lock-violation codes) becomes `BlockingIOError`; unsupported locking, invalid handles, and other permanent `OSError` values now propagate immediately. A regression replaces the lock primitive with `EINVAL` and asserts that no retry/sleep occurs.

The refactor also removes the temporary source-snapshot workflow, extracts key initialization out of the oversized store module, and keeps all fail-safe authority verification unchanged.

PR #2527, opened 2026-08-21
fix(guard): make local-CLI schema and update failures self-diagnosing

kantorcodes:
## Summary
- `ensure_local_cli_schema` now raises a typed `LocalCliSchemaError` (still a `ValueError` subclass, so existing handlers are unaffected) that separates two failure modes which previously shared one opaque message: a store written by a **newer Guard build** ("update Guard and retry") versus a schema marker matching no known build (damage or an unsupported build). Store opening stays fail-closed in both cases; only the remediation text differs. Every surface that prints the exception — daemon logs, CLI `Error:` output, desktop support panels — now names the recovery step.
- The trusted distribution probe (`update_subprocess`) now reports the version embedded in the installed **code files** (`codex_plugin_scanner.__version__`) next to the dist-info metadata version, with the same strict JSON-record validation as the existing fields.
- After an installer run, `hol-guard update` verifies those two versions agree. A partially applied install — fresh metadata beside stale code — now fails the update with reason `update_install_inconsistent` and a retry command, instead of being reported as a successful update. Such mixed installs previously passed the metadata-only version check and later surfaced as store schema crashes at startup, taking the daemon down with no actionable message.

## Testing
- `python -m pytest tests/test_guard_local_cli_schema.py` — newer-store marker raises the typed error with update guidance and correct version attributes; tampered and unknown markers keep the integrity message.
- `python -m pytest tests/test_guard_update_subprocess.py -k distribution_query` — probe accepts and normalizes `code_version`, rejects invalid values and unknown keys.
- `python -m pytest tests/test_guard_phase03_local_install.py -k "installed_code or code_disagrees"` — full update flow fails with `update_install_inconsistent`, retry command, `changed: false`; consistency helper unit cases for match, mismatch, and missing probe field.
- Full `tests/test_guard_update_subprocess.py`, `tests/test_guard_update_isolation.py`, `tests/test_guard_update_pipx_recovery.py`, `tests/test_guard_phase03_local_install.py`, `tests/test_dashboard_update.py`, `tests/test_guard_update_daemon_handoff.py` suites pass except two pre-existing install-layout-dependent tests that fail identically on the base commit under the same local harness.
- Ratchet baseline bumped for the 11 new cases and 199 added test lines.

## Notes
- No behavioral change to protection decisions: schema verification remains fail-closed; the updater gate only converts a silent partial install into a named, retryable failure.


kantorcodes:
All four review threads are addressed in 3ed6d652c:

- **Single post-install probe:** the version seam's implementation now delegates to one `verify_installed_distribution` call, so exactly one trusted subprocess serves both the resulting version and the consistency check (no TOCTOU between reads, no duplicate spawn).
- **No package import to read the version:** the probe reads the installed `version.py` text directly (bounded to 4 KiB) and the version assignment is parsed outside the subprocess — no install-time code executes during verification.
- **Import/parse failure no longer skips the check:** a version source that is missing, unreadable, or has no parseable `__version__` yields an empty code version, and empty fails verification as `update_install_inconsistent`. Only a probe from an older build without the field keeps the skip semantics.
- **Quality ratchet:** the probe script and validation moved to a new small module; `update_subprocess` shrinks by 65 lines and the baseline records the 11 added lines in `update_commands` plus the new test file.

PR #2529, opened 2026-08-22
adapters: add OptiQra harness support

armin5872:
# Add OptiQra as a HOL Guard harness (release/3.0)

Adds `optiqra` as a recognized HOL Guard harness so `/api/auto-fix-project`
can call `guard hook --harness optiqra` with a resolved fix action and its
target files, and get back an allow / deny / approval-required decision
before it writes anything to disk.

## What this does

- Registers `optiqra` as a harness (`adapters/optiqra.py`), matching the
  existing adapter interface (`HarnessAdapter`).
- Adds `normalize_optiqra_payload` so `guard hook --harness optiqra` can
  turn a resolved action into a typed `GuardActionEnvelope`, reusing the
  same generic normalizer every other harness (Codex, Kimi, etc.) uses —
  no bespoke parsing.
- Wires `optiqra` into the CLI's action-envelope harness set so the hook
  path actually builds an envelope for it.
- Registers a `HarnessProtectionContract` + protection-capability entry so
  `optiqra` shows up correctly in setup/coverage tooling and docs, and
  updates the generated docs table and product-model schema to match
  (both are checked by existing tests that assert they stay in sync with
  the contract registry).

## What this deliberately does NOT do

- **It does not evaluate the diff.** `GuardActionEnvelope` has no
  diff/content field. Guard gates on the *resolved action and target file
  paths* OptiQra sends, not on the diff text itself. If OptiQra puts diff
  content under a content-shaped payload key (e.g. `content`), Guard's
  existing redaction pipeline strips it before policy ever sees it —
  there's a test (`TestOptiQraDiffVisibilityLimitation`) that pins this
  down explicitly so it doesn't regress silently.
- It does not add local config discovery, install, or uninstall support.
  OptiQra has no on-disk config file or launchable executable the way
  Kimi/ZCode/Pi do — it's expected to call `guard hook --harness optiqra`
  directly from its own backend. `detect()` honestly reports "not locally
  installed" instead of inventing a config file to scan for, and
  `install()`/`uninstall()` raise `NotImplementedError` with a message
  pointing at the actual integration point, rather than silently
  installing a shim nothing will ever call.

If diff-level evaluation is something you want in this pass rather than a
follow-up, that's a small but real change to the shared
`GuardActionEnvelope` schema (a new explicit field, deliberately excluded
from the generic redaction sweep) — touches every harness, not just this
one, so I left it out of this PR pending your input.

## Architecture

```
OptiQra resolves fix action
  -> target files + (today: no diff visibility)
  -> OptiQra's backend calls: guard hook --harness optiqra
       (JSON payload on stdin: {"tool_name": "write_file",
        "tool_input": {"file_paths": [...]}})
  -> Guard normalizes payload -> GuardActionEnvelope
       (action_type="file_write", target_paths=[...])
  -> Guard evaluates against configured policy
  -> exit code + JSON on stdout:
       allow  -> rc 0,  policy_action: "allow"   -> OptiQra proceeds to write
       block  -> rc !=0, policy_action: "block"  -> OptiQra must not write
       review -> rc !=0, policy_action: "review" -> OptiQra must not write,
                                                      surface for approval
```

`allow`/`block`/`review` are exactly Guard's existing action vocabulary —
nothing OptiQra-specific was invented there.

## Handling the allow/deny/approval result

OptiQra's own backend is responsible for branching on the hook's exit code
and JSON output before it ever reaches the write step in
`/api/auto-fix-project`:

- exit code `0` → the write may proceed.
- non-zero exit code → the write must **not** proceed. The JSON
  `policy_action` field (`"block"` vs `"review"`) tells OptiQra which case
  it is, so it can either reject the fix outright or surface it through
  its own UI as pending approval — but either way, nothing gets written
  until a subsequent `allow` comes back.

This mirrors the existing convention (Kimi, ZCode, generic hook path):
Guard is fail-closed by default — an unparseable, malformed, or garbage
payload is still evaluated against the configured default policy action;
it is never treated as an implicit allow.

## Tests (`tests/test_optiqra_adapter.py`)

- Adapter identity (`harness`, `approval_tier`, registration/resolution).
- `detect()` reports the honest "not locally installed" state.
- `install()`/`uninstall()` raise `NotImplementedError` rather than
  silently no-op-ing or installing a meaningless shim.
- `normalize_optiqra_payload` maps a `write_file`-shaped payload to
  `action_type="file_write"` with the correct `target_paths`.
- A malformed/empty payload still normalizes (no crash) and is still
  evaluated against the configured policy — never silently allowed.
- Full hook-level flow via `_run_hook_generic_payload` for all three
  outcomes:
  - allow → rc `0`, `policy_action: "allow"`
  - block → rc `!=0`, `policy_action: "block"`
  - review → rc `!=0`, `policy_action: "review"` (distinct from block)
- Target files reach the recorded artifact/decision.
- The diff-redaction limitation, pinned as a test so it's visible if this
  ever silently changes.

18 tests, all passing.

## Test results

```
$ pytest tests/test_optiqra_adapter.py -v
...
============================== 18 passed in 8.27s ==============================
```

Also re-ran the suites most likely to be affected by adding a new harness
to the shared registries, all passing after fixing two checked-in
snapshots that needed to include the new harness (see below):

```
$ pytest tests/test_guard_runtime_action_harnesses.py tests/test_guard_harness_setup.py \
         tests/test_guard_harness_contracts.py tests/test_guard_phase04_harness_contracts.py \
         tests/test_guard_protection_capabilities.py tests/test_guard_product_model_contracts.py \
         tests/test_cline_completion_gates.py tests/test_optiqra_adapter.py -q
...
228 passed
```

Plus the full existing adapter suites (zcode, kimi, pi, hermes, grok,
openclaw) — 262 tests, unaffected. Full-repo test collection (15,322
tests) succeeds with no import errors, so nothing broke at load time
elsewhere in the tree.

## Diff review — everything touched and why

| File | Why |
|---|---|
| `src/codex_plugin_scanner/guard/adapters/optiqra.py` | New adapter (identity, honest `detect`, unsupported `install`/`uninstall`). |
| `src/codex_plugin_scanner/guard/adapters/__init__.py` | Registers the adapter (1 line). |
| `src/codex_plugin_scanner/guard/adapters/contracts.py` | Adds the `HarnessProtectionContract` + display name so setup/coverage tooling knows about `optiqra`. |
| `src/codex_plugin_scanner/guard/runtime/actions.py` | Adds `normalize_optiqra_payload`, registers it — reuses the existing generic normalizer, no new logic. |
| `src/codex_plugin_scanner/guard/cli/commands_support_hook_payload.py` | Adds `"optiqra"` to the set of harnesses the hook CLI will build an envelope for (1 line). |
| `src/codex_plugin_scanner/guard/protection_capabilities.py` | Adds the capability tuple for `optiqra`, matching the existing CLI-hook-boundary harnesses (Kimi/Grok/Pi/OMP/ZCode) — `fail_open_on_hook_failure=True` because, unlike a Guard-managed local shim, whether a failed hook call fails open is entirely controlled by OptiQra's own code, not Guard's. |
| `src/codex_plugin_scanner/guard/schemas/guard_product_model_v1.json` | Checked-in snapshot of `supported_harnesses`; regenerated to include `optiqra` (caught by an existing test). |
| `docs/guard/harness-support.md` | Generated contract table + a short prose section for `optiqra`, matching every other harness's doc entry; caught by an existing doc-sync test. |
| `tests/test_optiqra_adapter.py` | New test file (18 tests, see above). |

Nothing here touches OptiQra's own repository — the adapter side is
entirely in HOL Guard, as requested. Wiring OptiQra's
`/api/auto-fix-project` handler to actually call `guard hook --harness
optiqra` before its write step is a follow-up in OptiQra's own codebase,
not part of this PR.

## Known gap (flagging up front, not burying it)

This does **not** give Guard visibility into the actual diff content —
only the resolved action and target file paths. That was explicitly the
part you asked for ("hand Guard... the final diff"). Extending
`GuardActionEnvelope` with an explicit, non-redacted diff field is a small
but real schema change that affects every harness, so I left it out of
this "smallest possible adapter" PR pending your call on whether you want
it bundled in or as a fast follow-up.

### DIFF
--- a/docs/guard/harness-support.md
+++ b/docs/guard/harness-support.md
@@ -117,6 +117,12 @@ Current Guard support in this repo:
   - installs Guard-managed `PreToolUse` and `UserPromptSubmit` hooks in the `hooks` section of `~/.zcode/cli/config.json` without touching user `mcp`, `plugins`, or pre-existing hooks
   - blocks by returning exit code `2` and ZCode-native stdout JSON `hookSpecificOutput.permissionDecision: "deny"` with approval-center copy in stderr
   - fails open if a hook crashes or times out, so ZCode keeps working when Guard is unreachable
+- `optiqra`
+  - has no local config file or launchable executable for Guard to discover; `detect()` always reports not installed, and `install`/`uninstall` are unsupported since there is no local shim to manage
+  - is expected to call `guard hook --harness optiqra` directly from its own `/api/auto-fix-project` backend, once per resolved fix action, before that action is written
+  - the resolved action's target files are passed as `tool_input.file_paths` on a `write_file`-style payload, which Guard normalizes into a `file_write` action with those target paths
+  - Guard's action envelope has no dedicated diff/content field yet: any payload key that looks like file content (for example `content`) is redacted before policy evaluation, so today Guard gates on the resolved action and target files only, not the diff text itself
+  - like the other CLI-hook-boundary harnesses, whether a failed or unreachable `guard hook` call fails open is controlled by OptiQra's own calling code, not by Guard, since there is no Guard-managed shim in the loop
 
 Gemini, Antigravity, and shared Codex/AIBOM skill discovery bind approval and
 inventory identity to the complete accepted skill directory rather than only
@@ -188,4 +194,5 @@ Generated from `src/codex_plugin_scanner/guard/adapters/contracts.py`.
 | `grok` | `grok`, `grok-build`, `grok-build-cli`, `xai-grok` | ❌ | ✅ | ✅ | shell, prompt, mcp_tool, file_read, file_write |
 | `pi` | `pi`, `pi-agent`, `pi-coding-agent` | ✅ | ✅ | ✅ | shell, prompt, mcp_tool, file_read, tool_result |
 | `omp` | `omp`, `oh-my-pi` | ✅ | ✅ | ✅ | shell, prompt, mcp_tool, file_read, tool_result |
+| `optiqra` | `optiqra` | ❌ | ❌ | ❌ | file_write |
 | `zcode` | `zcode`, `zai`, `z-code`, `zai-zcode` | ❌ | ✅ | ❌ | shell, prompt, mcp_tool, file_read |
--- a/src/codex_plugin_scanner/guard/adapters/__init__.py
+++ b/src/codex_plugin_scanner/guard/adapters/__init__.py
@@ -22,6 +22,7 @@ _ADAPTER_SPECS: tuple[tuple[str, str], ...] = (
     (".pi", "PiHarnessAdapter"),
     (".openclaw", "OpenClawHarnessAdapter"),
     (".opencode", "OpenCodeHarnessAdapter"),
+    (".optiqra", "OptiQraHarnessAdapter"),
     (".zcode", "ZCodeHarnessAdapter"),
 )
 
--- a/src/codex_plugin_scanner/guard/adapters/contracts.py
+++ b/src/codex_plugin_scanner/guard/adapters/contracts.py
@@ -143,6 +143,7 @@ _DISPLAY_NAMES = {
     "pi": "Pi",
     "omp": "Oh My Pi",
     "zcode": "ZCode",
+    "optiqra": "OptiQra",
 }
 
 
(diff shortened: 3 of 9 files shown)


kantorcodes (changes requested):
18 focused tests pass, but the integration invokes the nonexistent `guard hook` CLI instead of the package-exported `hol-guard`/`plugin-guard`; approval copy falsely claims final diff evaluation even though content is redacted; `{}` normalizes to a targetless `config_change`, and a real `hol-guard hook` invocation exits 0 under the default `warn` policy; and setup contracts advertise connect/repair/install while the adapter raises `NotImplementedError`. Please require a fail-closed, validated write payload and an honest, actually supported setup contract before merge.

PR #2598, opened 2026-08-26
feat(rust): enforce hot-path authority and remove PostToolUse Python fallback

kantorcodes:
## Summary

Implements the T001-T100 Rust hot-path stability tranche for `release/3.0`.

- adds a versioned ownership manifest covering hook decisions, resident transport, packaging, and migration governance
- adds an always-selected, merge-base-aware authority gate with hard-expiring waivers and unmapped-path rejection
- classifies both source and destination paths for renames and copies
- adds privacy-safe backend receipts and asynchronously persisted aggregate route counters
- removes the Python scanner/evaluator fallback from the daemon `PostToolUse` path
- converts native failure, including `HOL_GUARD_NATIVE=off`, into a deterministic fail-safe block instead of changing decision engines
- hardens mixed documentation/source target handling so documentation relaxation cannot weaken source scanning
- adds real compiled-runtime integration for one-shot, authenticated resident, and all supported harness routes

## Ownership impact

`PostToolUse` is Rust-authoritative after this change. Python remains a bounded transport and harness-mapping layer and cannot substitute `HookReviewEngine`, `ContentScanner`, or `HookDecisionCache` when native evaluation fails.

`PreToolUse` is explicitly staged for the immediately following delivery. The manifest carries one short, hard-expiring waiver through 2026-09-02. It does not allow Python fallback on a Rust-owned surface.

There is no `strict` mode. Rust is the product default.

## Validation

The authority workflow runs entirely on GitHub-hosted runners for pull requests and enforces every selected manifest job:

- Rust workspace formatting, Clippy, tests, release build, and runtime self-test
- one-shot and authenticated resident integration
- all-harness `PostToolUse` integration through the production worker route
- native command-model and Rust/Python differential gates needed by staged `PreToolUse`
- resident transport fault tests and performance budgets
- native release contracts and installed static native-wheel integration

The all-harness proof reads measured aggregate counters from the production routing layer and requires zero Python decisions and zero native fail-safe outcomes. Evidence and reports reject both raw and JSON-escaped request content.

Targets `release/3.0` only. Never merge into `main`.


kantorcodes:
Local integration snapshot source: [release/3.0 source archive](https://github.com/hashgraph-online/hol-guard/archive/refs/heads/release/3.0.zip). This comment will be replaced with the final integration evidence after review remediation.

kantorcodes:
Unique-diff vs current `main`: this branch still carries route metrics, hotpath ownership TOML/scripts, all-harness integration, and data-plane path work that are not in `main`.

Supported PreToolUse/PostToolUse fail-closed `auto` behavior and stable native-wheel publication are now on #2652. Leaving this PR open until that unique evidence/metrics surface is either ported or explicitly dropped.

PR #2627, opened 2026-08-27
feat(rust): enforce native hot-path authority (T001-T100)

kantorcodes:
## Scope

Implements Rust migration tasks T001-T100 as the first delivery boundary on `release/3.0`.

- makes supported PostToolUse content review Rust-authoritative with deterministic fail-safe blocking when native evaluation is unavailable
- removes Python HookReviewEngine/ContentScanner/decision-cache fallback from the PostToolUse production decision path
- adds privacy-safe native route receipts and aggregate authority evidence
- adds a versioned hot-path ownership manifest and non-bypassable authority CI gate
- hardens multi-target source classification so documentation paths cannot relax source-code scanning
- adds real one-shot, authenticated-resident, transport-fault, all-harness, performance, and native-release integration gates
- upgrades the immutable Rust policy snapshot contract to v2 and installs generation-bound policy snapshots into the resident runtime
- moves policy/config loading out of the request hot path and binds resident enforcement posture to the installed snapshot

Rust is the product default. This PR does not introduce a `strict` runtime mode. PreToolUse authority is completed in T101-T200; the temporary manifest entry cannot permit Python fallback and has a hard expiry.

## Local integration proof

Executed against the compiled Rust runtime rather than mocks:

- Rust workspace tests/build for guard-hook-core, guard-policy-snapshot, and hol-guard-runtime
- `scripts/integration/rust_authority_integration.py` including authenticated resident framing faults
- `scripts/integration/rust_all_harness_integration.py`: 14 harnesses, 28 Rust decisions, 0 Python decisions, 0 native fail-safe outcomes
- Hook worker/native resident integration suite: 42 passed
- ownership validator self-test
- Ruff/compile checks on modified Python surfaces

Targets `release/3.0` only.


kantorcodes:
Unique-diff vs current `main`: this branch still carries `native_route_metrics`, all-harness/SLO integration scripts, and a PostToolUse-only worker shape that would drop the current `main` PreToolUse native path if merged as-is.

Supported fail-closed `auto` authority and stable native-wheel publication are now on #2652. Leaving this PR open for a follow-up that ports only the unique metrics/evidence pieces onto `main`.

PR #2673, opened 2026-08-30
feat(extensions): add MacScope MCP write-gate protection

rsm23:
## Summary

- Add the built-in `command.macscope` extension for the three MacScopeMCPServer write-enabling startup flags, with a distinct experimental-write permission.
- Preserve read-only and informational command classification, and keep combined write gates independently visible.
- Refresh the generated extension directory, catalog baselines, decision-diff report, and signed projection fixtures.
- Verify that help on one shell command does not suppress a write gate on a later command.

## Testing

- Focused MacScope, registry, permission, control-catalog, and generated-directory suite: 94 passed.
- `.venv/bin/python tests/guard_command_decision_diff.py --check`
- `.venv/bin/python -m pytest tests/test_guard_command_decision_diff.py --tb=short`: 6 passed.
- Signed projection vector test: 1 passed.
- Ruff on changed Python files and generated extension-directory check passed.
- Built and smoke-tested the local wheel: all three write gates and their combination require review; read-only, help, and version forms remain unmatched by the extension.

## Notes

The extension covers the documented MacScopeMCPServer startup flags. It does not classify individual MCP calls or MCP configuration writes. The outstanding executable-identity review remains unresolved.

Startup parsing and informational exits were checked against the pinned upstream source already referenced by the extension.


kantorcodes (changes requested):
The MacScope extension itself looks good. The remaining blocker is the intentional test-suite ratchet from the new coverage: CI is measuring 16,318 collected cases and 352,932 test-source lines, while `ci/test-suite-ratchet-baseline.json` is still at 16,313 / 352,808. Please update that reviewed baseline to the observed values and push it to this branch; once the rerun is green I can finish the review.

rsm23:
Thanks, I updated the reviewed ratchet baseline in `f500c1a60`. Since the latest main was merged after your review, the current inventory is `16,323` collected cases and `353,078` test-source lines, so I used those exact observed values. The ratchet and all 88 focused tests pass locally. The GitHub workflows are awaiting approval to run.

kantorcodes (changes requested):
The ratchet-only follow-up fixes the CI baseline, but it does not address the command-boundary blocker from the previous review. Exact current probes still show that only the literal top-level MacScopeMCPServer plus literal --allow-feature-writes is gated: exec MacScopeMCPServer --allow-feature-writes, an alternate executable name, and a variable-expanded write flag all return no_match. Because workspace-controlled shell text can reach those equivalent startup forms, write-enabled MacScope can still start without review. Please normalize wrappers/expansions or enforce the write gate at the actual MCP action boundary, and add adversarial coverage for those forms.

kantorcodes (dismissed):
This PR predates the current Extension contribution architecture and is not safe to approve as-is. It is hundreds of commits behind current `main` and does not use the modern external contribution contract, so its runtime rules, catalog metadata, packaging, publisher attribution, generated identities, and control-state behavior cannot be assumed to match current Guard semantics.

Please rebase/reconstruct this contribution on current `main` using the current Extension kit: add a `contributions/extensions/command.macscope.json` manifest, classify it as external/opt-in in the trust map, include the manifest in both wheel and frozen-runtime packaging, register only the intended current structured matchers, and regenerate the directory/catalog baseline plus managed projection identity from the final tree. Because MacScope can perform destructive/root-required macOS operations, add current inspection/runtime tests for enabled/disabled control state, safe/read-only counterparts, wrapper/option/expansion ambiguity, fail-secure malformed input, and evidence privacy (no raw paths/arguments/secrets persisted). Then rerun the current registry/trust/package/type/policy checks and request re-review.

kantorcodes (changes requested):
This needs to be rebuilt on current `main` using the current external Extension contract: contribution manifest, external/opt-in trust entry, wheel + frozen packaging, current generated identities, and fail-secure wrapper/expansion tests. The current branch is too old to merge safely.

PR #2674, opened 2026-08-30
feat(guard): add essh remote execution and cache removal extension

matthart1983:
## Summary

- Add essh host-group execution and cached-entry removal protection as an external extension, disabled until explicitly enabled by a local administrator.
- Register the contribution with a dedicated detector module and community publisher metadata.
- Handle interleaved global options and local help/version exits while preserving review of remote arguments after `--`.
- Refresh generated catalog artifacts and extract the unchanged action-risk map to keep the rule catalog within the quality limit.

## Testing

- Focused remote, matcher, trust, and catalog tests.
- Contribution-schema validation and default-off, enabled, and disabled-again regressions.
- Generated directory, catalog/projection fixtures, decision-diff checks, Ruff, and structural quality checks.
- Isolated wheel checks for packaged contribution metadata and activation behavior.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in Guard coverage for the `essh` remote-administration tool.
  - Reviews `essh run` commands executed across host groups.
  - Reviews removal of cached hosts, keys, and workspaces.
  - Classifies group execution as critical risk and cache removal as high risk.
  - The extension is disabled by default and requires explicit local enablement.

- **Documentation**
  - Added `essh` coverage details, command examples, risk guidance, and official reference information to the Guard documentation.


matthart1983:
Thanks — the interleaved global option finding was correct and reachable, not theoretical.

`--theme` is a clap `global = true` argument in essh, so it is accepted between subcommand levels. I confirmed against the built binary that `essh hosts --theme dark remove --help` parses, which means `essh hosts --theme dark remove web-1` removes the host while evading `command.remote.essh.cache-removal`. The cause is that `LeadingSubcommandMatcher` stops consuming options at the first operand and then compares the subcommand chain positionally, so `("hosts", "--theme")` never matched `("hosts", "remove")`.

Fixed in c24a21d by adding an opt-in `interleaved_options_with_values` field to `LeadingSubcommandMatcher`, which skips declared options — separated and `=` forms — while walking the subcommand chain. It defaults to empty, and with an empty set the walk is equivalent to the previous positional slice, so the redis, rabbitmq, and NATS callers are unaffected. There is a regression test pinning that equivalence in both directions.

Added coverage for `hosts`, `keys`, and `workspace` removal with `--theme` before, between, and after the subcommands, plus `essh run --theme dark web -- uptime`. Safe cases assert no false positives on `essh hosts --theme dark list`, `--theme dark add`, and `workspace --theme dark show`.

Worth flagging that this pattern is not essh-specific: any multi-level CLI with global options has the same shape, so other extensions may want the new field too. I have left those alone in this PR.

kantorcodes (changes requested):
Two blockers remain before this can land. First, please address the still-open Qodo correctness finding so top-level essh `-h`/`--help`/`-V`/`--version` exits stay safe for both `run` and cache-removal paths, while a literal `--help` after the remote-command `--` delimiter still counts as remote execution. Second, CI confirms the new test is intentional ratchet growth: 16,314 collected cases and 352,915 test-source lines versus the current 16,313 / 352,808 baseline, so please update `ci/test-suite-ratchet-baseline.json` to those observed values after the correctness-test changes (using the final post-fix inventory if it increases further). The rest of the refreshed checks I inspected are passing.

matthart1983:
Both blockers addressed in 9de8e14. Thanks for the merge from main and the precise ratchet numbers.

**1. Help and version exits**

Confirmed against the essh binary: it inherits clap's `-h`/`--help`/`-V`/`--version`, which print and exit without running the subcommand. Four invocations were being raised for review despite executing nothing — `essh --help run`, `essh --help keys remove deploy-key`, `essh run --help`, and `essh hosts remove --help`.

I went slightly wider than top-level position. `essh run --help` and `essh hosts remove --help` put the flag after the subcommand rather than before it, and `forbidden_flags` only inspects leading flags, so those two would have stayed false positives. They are equally non-executing, so the fix covers the whole region ahead of the delimiter.

Added an opt-in `forbidden_flags_before_delimiter` field to `LeadingSubcommandMatcher` that suppresses a match when a declared exit-only flag appears before `--`, in long, short, and clustered short forms. **The scan stops at `--`**, so `essh run web -- --help` is still remote execution and still matches, as you asked. `essh run web -- essh --help` also still matches. Defaults to empty, so redis, rabbitmq, and NATS are unchanged; a regression test pins both directions.

**2. Ratchet baseline**

Updated to the post-fix inventory rather than the values you observed, since the new correctness tests grew it further, per your note. Measured with `scripts/ci/test_inventory.py`:

- `maximum_collected_cases`: 16313 → **16315**
- `maximum_test_source_lines`: 352808 → **352951**

`scripts/ci/test_suite_ratchet.py` passes against it locally.

**On the three red checks**

All three traced to the one ratchet failure. `quality` and `contract` both failed on the ratchet directly, and `ci (3.12)` is a gate job that only asserts the others succeeded — its log shows `QUALITY_RESULT: failure` with `TESTS_RESULT: success`, so the full suite itself was green on 3.12.

Also regenerated the decision-diff source bindings again, since this touches two source files.

kantorcodes (commented):
Latest head addresses the ESSH parser/help-version cases from the prior review. The remaining blocker is branch drift: `main` advanced and GitHub's update-branch operation now reports merge conflicts, while the maintainer integration cannot write into the fork head directly. Please merge/rebase current `main` into this branch and regenerate the derived ratchet/catalog/projection fixtures from that refreshed tree rather than preserving the pre-merge digest values. I’ll re-check the refreshed head and CI from the PR.

kantorcodes (approved):

PR #2686, opened 2026-08-30
fix(guard): fail closed on unbound native commands

shayansal:
## Summary

- restrict native automatic authorization to argument-free POSIX shell built-ins
- require review for external commands until the native request binds the resolved executable identity and argument effects
- add unit and real-binary regression coverage for executable-path spoofing, process-launching helper options, environment-selected helpers, and output-path writes
- keep the ordinary prompt-free boundary test on the bounded `pwd` case

## Security rationale

The v1 native PreToolUse request does not carry enough execution context to prove which external executable the harness will resolve or whether its arguments can launch helpers or write outside the workspace. Treating those commands as explicitly benign therefore creates a policy bypass for an in-scope malicious workspace.

This patch deliberately fails closed: external Git/search/metadata commands now require review until executable identity and per-command argument effects are cryptographically or structurally bound to the native decision.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --locked --workspace --all-targets -- -D warnings`
- `cargo test --locked --workspace --all-targets` — 39 passed
- native authority ownership and no-Python-gate checks
- release runtime build and self-test
- expanded real-binary PreToolUse adversarial integration suite
- Ruff check/format for the changed Python test
- `git diff --check`

This is opened as a draft so maintainers can coordinate the security handling and release plan before it is marked ready.

## Summary by CodeRabbit

- **Bug Fixes**
  - Tightened automatic command approvals to allow only argument-free `pwd` and `true` shell built-ins.
  - Commands involving sensitive text, environment variables, redirection, path-prefixed executables, or risky Git and search options now require review or are denied.
  - Improved fail-closed handling for potentially unsafe command patterns.
- **Tests**
  - Expanded coverage for adversarial command scenarios and updated safe-command checks.


kantorcodes (changes requested):
bare pwd/true remain unbound to eventual shell/startup environment, so imported functions/BASH_ENV can execute after explicit allow; scripts/integration/rust_pretool_adversarial.py still expects whoami/uname/git/rg/grep/stat allow and will deterministically fail its rust-triggered workflow. Local cargo guard-command 11 and authoritative boundary pytest 22 passed but do not cover these blockers.

PR #2710, opened 2026-08-31
feat(guard): add blitcp command safety rules

gekap:
Follows up on the discussion about covering [blitcp](https://github.com/gekap/blitcp) — a file copier that moves data to local disks, SSH hosts, and object storage. I maintain it, so the flag semantics here are first-hand rather than inferred from the help text.

## The direction problem

blitcp's grammar is `SOURCE... DESTINATION`. The final operand decides whether a run sends data off the host or pulls it down:

```
blitcp /data s3://backups/nightly     # egress — local data leaves the host
blitcp s3://backups/nightly /restore  # a read — nothing leaves
```

Matching *any* operand would prompt on both, which trains people to click through the one that matters. No existing matcher expresses "last operand only", so this adds `TrailingOperandPrefixMatcher` to `command_structured_matchers.py`. It removes options and their values, then tests only `operands[-1]` against the given prefixes. It carries `required_flags` and `forbidden_flags` so a preview can be told apart from a real copy.

It is not an `AnyMatcher`, so it cannot go through `safe_flag_variant()`; the `--dry-run` variant is the same matcher built with `dataclasses.replace(..., required_flags={"--dry-run"})`, which is what that helper produces for the others.

## Rules

| rule | severity | trigger |
|---|---|---|
| `command.blitcp.remote-destination` | high, `network_egress` | final operand is `s3://`, `az://`, `gs://` or `smb://`; `--dry-run` is a declared safe variant |
| `command.blitcp.privilege-escalation` | critical, `execution` | `--use-sudo` re-runs the copier under sudo |
| `command.blitcp.self-update` | high | `--update` downloads a release and replaces the running executable in place |
| `command.blitcp.unverified-copy` | medium | `--no-verify` skips the post-copy verification phase |

On `--no-verify`, to be precise about what it does: blitcp's last phase reads every copied file back and hashes it against the source. `--no-verify` skips that phase, so the run reports success without having read anything back. The result is a silently incomplete copy rather than a destructive action, hence medium. It is not matched on a dry run, which copies nothing and therefore verifies nothing.

**On scope:** you asked for a risk-relevant subset and named destination operands, `--dry-run` and `--no-verify`. I included `--use-sudo` and `--update` as well, on the grounds that re-executing under sudo and overwriting the running binary from the network are the other two things blitcp does that change the trust picture. If you would rather keep this PR to the three you named, say so and I will drop those two commits' worth of rules — no argument from me.

## Options that consume a value

All 30 of them are declared, extracted from blitcp's own argparse rather than from `--help`. This is load-bearing: without `--s3-region` in the set,

```
blitcp /data s3://bucket --s3-region eu-west-1
```

reads `eu-west-1` as the destination and the upload stops matching entirely. There is a regression test for that shape and for four others like it.

## Pinned values that moved — please check the last one

Adding built-in rules moves the catalog identity, so five pins are refreshed:

- rule-count assertion in `test_guard_command_extensions.py` (146 → 150)
- `catalog_digest` in `test_guard_command_permission_catalog.py`
- the generated `tests/fixtures/extension-controls/catalog-baseline.v1.json`
- `tests/fixtures/guard-command-corpus/decision-diff-report.json`, regenerated with its own `--write`
- `contracts/managed-controls/v1/extension-projection-digest-vector.json`

Counts move 60 → 61 extensions, 146 → 150 rules, 158 → 162 permissions, which matches four new rules in one new extension exactly.

The last one deserves your attention: it is `guard.signed-cloud-extension-projection-vector.v1`, a contract the docs describe as shared between Local and Cloud. I regenerated it from the code because the test is otherwise red, but **if the cloud side pins the same digest it needs updating in lockstep** — I have no way to check that from here. This is not specific to this PR: any change to the shipped rule catalog will move that digest, so it may be worth a note in CONTRIBUTING either way.

## Tests

`tests/test_guard_command_blitcp_extensions.py` mirrors the remote extension tests: 11 review cases, 17 safe cases, and unit tests on the matcher. Both directions are covered explicitly, including `blitcp /data s3://bucket` matching while `blitcp s3://bucket/data /local/restore` does not.

Also covered: upper-case schemes stay unmatched (blitcp itself only accepts lower-case, so flagging `S3://` would prompt on a command blitcp refuses to run), a lone remote operand is not a copy, a scheme has to start the operand rather than appear inside it, and `--dry-run` before or after the operands is safe.

Checks:

- `uv run python -m ruff check src tests` — clean
- `uv run python -m ruff format --check` — clean on every file touched here
- `uv run pytest` — full suite, 16176 passed / 56 failed / 6 errors

On those failures: I ran the failing set against unmodified `main` and diffed. 43 fail identically without this branch, 17 are artifacts of running with `-n 12` and pass in isolation either way, and the 2 that were genuinely mine were the baseline pins above, now fixed. Re-running the comparison after the fix leaves the set of failures attributable to this branch empty. Happy to paste the per-test breakdown if useful.

## Summary by CodeRabbit

* **New Features**
  * Added safety detection for `blitcp` copy commands.
  * Classifies remote destinations, privilege escalation, self-updates, and unverified copies with appropriate risk levels.
  * Supports S3, Azure, Google Cloud, SMB, SSH-style destinations, and saved remote connections.
  * Handles dry runs, previews, remote-source reads, local copies, and Windows drive paths appropriately.
* **Tests**
  * Added comprehensive command-matching and risk-classification coverage.
* **Maintenance**
  * Updated catalog and release digests for expanded protection coverage.


gekap:
Good catch on the SSH destinations — that was a real gap, and an embarrassing one given SSH tar streaming is blitcp's oldest remote transport. `blitcp /data user@host:/backup` was leaving the host with no prompt at all. Fixed in 851c9a2.

`[user@]host:path` is not a prefix, so rather than stretch the prefix matcher I added `TrailingOperandHostTargetMatcher` next to it and composed both into the one rule. Both syntaxes mean the same thing — data leaving the host — and two separate rules would imply only object stores are worth reviewing. Direction is read identically, so a restore still does not match.

Two exclusions are load-bearing, both with tests:

- **A single-letter host is a Windows drive, not a remote.** `blitcp /data C:\backup` has to stay silent. blitcp's own path parser makes this distinction; without it every copy to a Windows drive prompts as egress, which is exactly the noise that trains people to click through the prompt that matters.
- **An operand containing `://` belongs to the prefix matcher**, so the two never both fire on one command.

The grammar is parsed by hand rather than by regular expression — it is small, and a matcher sitting on the review path should not carry backtracking behaviour that scales with operand length.

Behaviour now:

| command | result |
|---|---|
| `blitcp /data user@host:/backup` | review |
| `blitcp /data host:/backup` | review |
| `blitcp /data user@[2001:db8::1]:/backup` | review |
| `blitcp /data user@host:/backup --ssh-dst-port 2222` | review |
| `blitcp user@host:/backup /local/restore` | allow |
| `blitcp --dry-run /data user@host:/backup` | allow |
| `blitcp /data C:\backup` | allow |

Catalog identity moved again, so the affected pins are refreshed in the same commit — including the shared local/cloud projection vector I flagged in the description, which is still worth a look from someone who can see the cloud side.

kantorcodes:
> Good catch on the SSH destinations — that was a real gap, and an embarrassing one given SSH tar streaming is blitcp's oldest remote transport. `blitcp /data user@host:/backup` was leaving the host with no prompt at all. Fixed in [851c9a2](https://github.com/hashgraph-online/hol-guard/commit/851c9a248acb9156ee26716aa7c229735d93ba89).
> 
> `[user@]host:path` is not a prefix, so rather than stretch the prefix matcher I added `TrailingOperandHostTargetMatcher` next to it and composed both into the one rule. Both syntaxes mean the same thing — data leaving the host — and two separate rules would imply only object stores are worth reviewing. Direction is read identically, so a restore still does not match.
> 
> Two exclusions are load-bearing, both with tests:
> 
> * **A single-letter host is a Windows drive, not a remote.** `blitcp /data C:\backup` has to stay silent. blitcp's own path parser makes this distinction; without it every copy to a Windows drive prompts as egress, which is exactly the noise that trains people to click through the prompt that matters.
> * **An operand containing `://` belongs to the prefix matcher**, so the two never both fire on one command.
> 
> The grammar is parsed by hand rather than by regular expression — it is small, and a matcher sitting on the review path should not carry backtracking behaviour that scales with operand length.
> 
> Behaviour now:
> 
> command	result
> `blitcp /data user@host:/backup`	review
> `blitcp /data host:/backup`	review
> `blitcp /data user@[2001:db8::1]:/backup`	review
> `blitcp /data user@host:/backup --ssh-dst-port 2222`	review
> `blitcp user@host:/backup /local/restore`	allow
> `blitcp --dry-run /data user@host:/backup`	allow
> `blitcp /data C:\backup`	allow
> Catalog identity moved again, so the affected pins are refreshed in the same commit — including the shared local/cloud projection vector I flagged in the description, which is still worth a look from someone who can see the cloud side.

This is looking great! Mind fixing the conflicts in the branch?

kantorcodes (changes requested):
Two concrete blockers remain before this is ready to merge:

1. The remote-destination matcher still misses blitcp's saved connection destinations (`NAME` / `NAME:subpath`), so a copy to a saved cloud connection can bypass the egress review even though literal `s3://` / `az://` / `gs://` / `smb://` destinations are covered. Please add structured coverage for those destination forms plus regressions that preserve remote-source reads as safe.
2. The flag-only `--no-verify` / `--use-sudo` matchers still fire without proving there is a copy operation, e.g. help/bare invocations. Please add operand/copy-shape validation (or equivalent safe handling) and regressions so those flags only elevate an actual blitcp copy.

The shared Local/Cloud digest vector also moved in this PR; I’ll treat that as a maintainer-side rollout check rather than asking you to invent a cross-repo change here.

gekap:
Both blockers addressed, and the branch is now conflict-free against main.

**Saved connections** — `TrailingOperandRemoteAliasMatcher` mirrors blitcp's own `resolve_named_endpoint` exclusion for exclusion, so `blitcp /data azure-prod:backups` (and `gcs:` with an empty subpath) now reviews as egress while the same alias as a source stays a silent restore. The bare `NAME` form is covered when `--credentials-file` appears on the command. Without that flag a bare name is statically indistinguishable from a local directory — prompting there would be the any-operand noise you agreed this PR was right to dismiss — so that residual (bare name + implicit default credentials file) is documented in the matcher rather than closed at the cost of prompting on every copy to a fresh local directory. If you'd rather close it unconditionally, it's a one-line flip (`allow_bare_names=True` without the flag gate) and I'll defer to you on the noise trade-off.

**Flag-only rules** — `--use-sudo` and `--no-verify` now require the shape of a copy (source + destination once option values are stripped) via `OperandGatedFlagMatcher`; bare/help invocations no longer prompt. Regressions cover both.

Also picked up CodeRabbit's point while in there: blitcp's `argv[1]` subcommands (`creds`, `ls`, `list-objects`, `deps`, …) are excluded from all blitcp matchers — `blitcp ls s3://bucket` is a read and no longer reviews as egress. The exclusion is deliberately case-sensitive on the literal first argument, because that is exactly how the tool dispatches.

Pins and the shared vector are regenerated on top of today's main; thanks for taking the cloud-side rollout of the vector. Full suite run locally against the same failing node ids on a clean main checkout: zero regressions attributable to this branch.

kantorcodes (approved):
Both blockers from the earlier review are closed on the current head, verified by direct inspection probes:

- Saved-connection destinations: `NAME:subpath` always reviews as remote egress; a bare `NAME` reviews when `--credentials-file` puts saved connections in play on the command; remote sources, plain local copies, and dry runs stay unreviewed (regression cases cover each). The residual bare-name gap — blitcp's implicit credentials file is auto-loaded from its own install directory — cannot be resolved by a pure command matcher and is documented in the extension.
- Flag rules are gated on the shape of a copy: `blitcp --use-sudo`, `--no-verify --help`, and single-operand invocations stay unreviewed while real copies with those flags review.

Additional fixes landed on the branch during review:

- `blitcp creds lock` / `creds unlock --use-sudo` — blitcp's documented sudo re-exec for credentials-file immutability — now reviews as privilege escalation; the plain forms stay unreviewed.
- Operand-direction matchers split into their own module so `command_structured_matchers.py` stays under the code-quality size limit; the shared rule builder is reused instead of duplicated; blitcp action risk classes merge beside its rules.
- The dry-run variant construction is typed for basedpyright, the credentials-gated alias matcher is narrowed to bare names so one operand yields one evidence row, and the unverified-copy description matches its destructive-shell classification (same convention as other silent-data-loss rules).

Catalog vectors and pins are regenerated for the new extension. CI, Desktop contract CI, Security Gates, CodeQL, Fuzzing, Rust authority, and I/O ownership pass on the final head; the Native wheel soak failure reproduces on `main` and is unrelated to this diff.

PR #2714, opened 2026-08-31
fix(guard): reload rotated credentials between refresh retry attempts

kantorcodes:
## Follow-up to #2712

CodeRabbit on #2712 flagged a Major concern: the refresh retry wrapper retries with the credentials captured *before* the first attempt, so if a peer process rotated the refresh token in between, the retry uses the stale token again — forcing the outer resolver's reload path to do the recovery work it was already doing, plus an extra HTTP round trip.

## Fix

Thread a `credential_reloader` callback into `_refresh_guard_oauth_access_token`. After the first invalid_grant, before the retry, the wrapper asks the caller for the latest stored credentials; the caller (`_resolve_guard_sync_auth_context_from_oauth_credentials`) re-reads `store.get_oauth_local_credentials` and returns the current refresh token + DPoP key material.

- The reloader runs only on the retry leg; the first attempt is unchanged.
- No behaviour change when the reloader returns `None` (missing credentials) — the retry falls back to the original token, preserving existing single-process semantics.

## Tests

New case in `tests/test_guard_oauth_refresh_resilience.py`:
- `test_invalid_grant_retry_uses_reloaded_credentials` — simulates a peer-process rotation between attempts and asserts the second request carries the rotated refresh token.

## Verification

- `pytest tests/test_guard_oauth_refresh_resilience.py tests/test_guard_oauth_reconnect.py tests/test_guard_oauth_client.py tests/test_guard_oauth_device_connect.py tests/test_guard_command_decision_diff.py` → 96 passed
- `ruff check` + `ruff format --check` clean on touched files
- Baselines bumped: runner.py oversized-file allowance 6512 → 6532; test-suite ratchet +1 case / +45 lines; decision-diff fixture re-hashed after runner.py change

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OAuth token refresh reliability when credentials are rotated concurrently.
  - Automatically retries with the latest available refresh token and key material after an invalid-grant error.
  - Persists and returns the credential set that successfully completes authentication, including the latest profile information.

- **Tests**
  - Added coverage for recovery using reloaded credentials and verifying persisted results.
  - Updated quality and test-suite baselines to reflect current project metrics.
  - Refreshed test fixtures to match current validation data.


kantorcodes:
Qodo's Major (Reloaded dpop key discarded) is now fixed in commits 9d17d3d41 and 1415172e3:

1. The wrapper now exposes `_effective_refresh_token` + `_effective_dpop_key_material` on the returned payload, reflecting whichever credential snapshot actually succeeded.
2. The reloader captures the reloaded credentials dict into a box ref, and `_persist_rotated_oauth_refresh_token` uses that reloaded dict — so on-disk state matches what the retry used.
3. The resolver's returned context's `dpop_key_material` reflects the snapshot that succeeded, so subsequent signed sync requests use the matching key.
4. Internal markers are popped before any downstream read so they never leak into persistence or assertions.

Added `test_invalid_grant_retry_persists_and_returns_effective_credentials` (in tests/test_guard_oauth_refresh_resilience.py) which forces a peer-rotation between attempts and asserts both persistence and the returned context use the rotated credential pair.

PR #2768, opened 2026-09-04
fix(release): stage every wheel force-include before PyInstaller freezes

kantorcodes:
## Summary
- Stage every `[tool.hatch.build.targets.wheel.force-include]` mapping into the source tree before PyInstaller freezes, instead of only the four Guard Cloud Review artifacts.
- Core releases that carry the extension trust-class map froze into binaries whose packaged resource lookup failed, fell back to a repository-relative path, and crashed before reporting their version — so recent stable releases published wheels without the signed desktop-core sidecars.
- Tests now cover generic mapping staging, missing-source fail-closed behavior, and that this repository's mappings include the extension trust-class map.

## Testing
- `uv run --no-sync pytest tests/test_stage_guard_cloud_review_artifacts.py`

### DIFF
--- a/scripts/release/stage_guard_cloud_review_artifacts.py
+++ b/scripts/release/stage_guard_cloud_review_artifacts.py
@@ -4,9 +4,13 @@ from __future__ import annotations
 
 import argparse
 import shutil
-import tomllib
 from pathlib import Path
 from typing import Any
+
+try:
+    import tomllib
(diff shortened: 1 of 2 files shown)


kantorcodes:
Superseded: main now stages the full artifact set and package initializers before the freeze; closing in favor of that fix.

PR #2818, opened 2026-09-06
feat(guard): add rungs command classification extension

ThroughTheWind:
## Summary

Adds `command.rungs`, an external opt-in command safety extension for the [rungs](https://github.com/ThroughTheWind/rungs) CLI, so users can apply different policies to repository inspection and repository-writing rungs commands. Classification lives entirely in Guard; rungs exposes no new metadata, as agreed with the maintainer.

Behaviour was verified from rungs source at git tag `v0.4.0` (commit `f7b6aa08466da08f5b0d495f9c9d62cf900cd8a7`, the version published to npm on 2026-09-05) and re-read at `3339065170f8cddb83c276f7044bdf22e95b9eb8` (48 commits later, 2026-09-06), where the dispatch of these four commands is unchanged. The relevant parser facts: the command is `argv[2]` with nothing before it (a switch there is an unknown command: help, exit 1, nothing runs); every later `--` token is a switch wherever it appears and `--name=value` registers the bare `--name`; only `--set` consumes a value; `--help`/`-h` are recognised only in the command slot, so `rungs check --help` runs the gates.

## Classification

| Command | Rule | Effect classes | Preview |
| --- | --- | --- | --- |
| `rungs doctor [path] [--explain]` | none (deliberately unmatched) | inspection: reads the tree, the install record and the ledger; `--explain` runs detectors in-process, skips `command` gates and only queries git read-only | n/a |
| `rungs check [path] [tier]` | `command.rungs.check` (high) | `execution` (every registered `kind = "command"` gate runs via `execSync`, commands the repository owns) + `destructive_shell` (appends to `.ai/.gate-ledger.jsonl`) | none: `--dry-run` is not read |
| `rungs add <module…>` | `command.rungs.add` (high) | `destructive_shell` (module files, gate registry, install record, rendered harness files) | `--dry-run` safe variant, verified: every write site is skipped |
| `rungs upgrade --apply` | `command.rungs.upgrade` (high) | `destructive_shell` (stale/missing files, gate blocks, record) | bare `rungs upgrade` plans only; `--dry-run` is ignored, so `--apply --dry-run` still reviews |

Repository inspection is expressed the way Guard already expresses it for `blitcp ls` and `repo2nb reverse` without `--force`: as an absence of review evidence, pinned by tests (with and without `--explain`, with stray switches, with positionals that spell other commands). No benign or observer exemption is added, so `npx @rungs/cli doctor` is not exempted by anything here.

Matching follows the skill-sunset precedent and targets the installed CLI (`rungs`, `rungs.cmd`, `rungs.exe`, path-qualified). The npm launchers (`npx`, `npm exec`, `bunx`) keep their existing package-firewall treatment; a test asserts the package intent is unchanged and that no rungs rule fires through them. `init`, `eject` and the other commands stay outside this boundary and keep Guard's fallback.

### Fixed command slot

`ExecutableMatcher` gains `require_leading_subcommands`. rungs reads its command from `argv[2]` alone, so the matchers require `check`, `add` or `upgrade` in that slot before any option logic runs; `rungs --full check` and `rungs --dry-run add backlog` no longer review, because rungs refuses them without running anything. Interspersed switches and the fail-secure parse still apply to everything after the command word, which is where forged previews live, and the safe-variant clone carries the requirement through. A matcher-level test covers the requirement, its interaction with unknown options, and the cloned dry-run variant.

Conservative choices worth knowing:

- `--dry-run` on `add` uses Guard's fail-secure parse: `--dry-run=false`, `--set --dry-run` (value swallowed), and `--mystery --dry-run` (unknown option that may swallow the switch) all review even though rungs 0.4.0 would preview two of them.
- `rungs upgrade --apply=false` reviews, because rungs registers the bare name of any assigned switch and applies. `--apply` is declared value-taking for that reason (comment in the module explains the mechanism).
- rungs' documented switches are declared so `rungs add x --into ../repo --dry-run` and `--confirm-paradigm --dry-run` stay previews.

## Boundary: classification, not recovery

rungs writes in phases with no journal, so a failed `add` or `upgrade --apply` can leave earlier files written (rungs tracks this as WI-098, open at the revision above). This PR classifies what an invocation can do before it runs. It does not detect a partial write, roll one back, or repair an installation; interrupted-write recovery is separate work in rungs.

## Files

- `src/codex_plugin_scanner/guard/runtime/command_rungs_extensions.py`: rules, matchers, risk map, spec (source references in comments)
- `src/codex_plugin_scanner/guard/runtime/command_rules.py`, `command_extension_matchers.py`: the `require_leading_subcommands` matcher option, threaded through the helper and the safe-variant clone
- `contributions/extensions/command.rungs.json`, `contracts/extensions/trust-class-map.v1.json`: external opt-in registration; `scripts/release/stage_guard_cloud_review_artifacts.py` stages the contribution file
- `command_builtin_extension_catalog.py`, `command_action_risk_classes.py`: registration
- `tests/test_guard_command_rungs_extensions.py`: review cases through the production path with a local-admin enable layer, inert/disabled states, previews and inspection (also proven with the extension enabled, so the safe assertions are not vacuous), leading-switch rejections, forged previews, wrappers and compound commands, launchers, out-of-scope commands, the matcher option, metadata
- Regenerated: extension directory README, catalog baseline fixture, projection digest vector, decision-diff report; pinned rule count (209 → 212 after merging `main`) and digests in three tests

## Testing

- `uv run pytest -q tests/test_guard_command_rungs_extensions.py` — 9 passed
- Regression suites for the other external extensions, the registry, directory, trust, contribution, catalog-detail, permission-catalog, policy-bundle, decision-routing, command-model, path-set-matcher, permission-examples and extension-builder suites together with the rungs suite — 306 passed
- `test_github_pr_create_body_file_honors_proposal_permission_toggle` (3 cases in `tests/test_guard_command_extensions.py`) fails identically on a pristine checkout of `main` in this environment (Windows), so it is a baseline failure unrelated to this change
- `uv run python scripts/render_command_extension_directory.py --check` — pass
- `uv run python -m ruff check src tests` — findings only in test files this PR does not touch, identical to pristine `main`
- `uv run python -m ruff format --check src tests` — 44 files would be reformatted, identical set on pristine `main`; every file in this PR is formatted
- `basedpyright --level error` on the changed modules and test — 0 errors
- `uv run pytest --tb=short` — on Windows the exact command stops at collection because `tests/guard_daemon_acceptance_fixtures.py` imports the Unix-only `resource` module; the result of the run with `--continue-on-collection-errors` is reported in the final comment

## Notes

- `origin/main` was merged in; conflicts were confined to the regenerated catalog identity artifacts, which were taken from `main` and regenerated on the merged tree.
- The decision-diff report was regenerated with POSIX-style relative paths for the source bindings: `tests/guard_command_decision_diff.py` hashes `str(path.relative_to(REPO_ROOT))`, which yields backslash ids on Windows, so a plain `--write` there produces a report CI would reject. The committed report matches what the generator produces on Linux (only the touched source hashes changed, as in previous extension additions).
- Not covered yet, by design: launcher forms (`npx @rungs/cli add …`), version-pinned package specs, `node src/cli.ts …` source checkouts, and the `init`/`eject` commands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added Guard support for the `command.rungs` extension.
  * Reviews Rungs commands that execute gates or modify repository files.
  * Supports previews, dry runs, and apply-required operations.
  * Added extension metadata and documentation to the Guard directory.

* **Bug Fixes**
  * Updated extension catalogs and trust metadata to include Rungs rules.
  * Improved command matching for required subcommand placement.

* **Tests**
  * Added coverage for matching, wrappers, launchers, safe commands, and risk classifications.


kantorcodes (dismissed):
The Rungs matcher fix is in, but the managed-controls projection is still internally inconsistent with the runtime extension catalog: the projection vector carries a different catalog digest than the registry/baseline it is supposed to describe. That is a release-contract correctness failure, not generated-file noise.

Please regenerate `extension-projection-digest-vector.json` from the current runtime catalog identity and update the paired delivery constants/tests from that regenerated vector. Then rerun the focused Rungs/extension registry plus policy-bundle delivery checks. I will re-review the regenerated head rather than approving mismatched identities.

kantorcodes (changes requested):
The managed-controls projection digest does not match the runtime extension catalog identity. Regenerate `extension-projection-digest-vector.json` from the current catalog and update the paired delivery constants/tests from that regenerated vector.

PR #2843, opened 2026-09-07
feat(guard): add tether memory MCP server contribution

sidyellur:
## Summary

Adds [tether](https://github.com/sidyellur/tether) (`tether-memory` on PyPI) as an External, opt-in MCP server contribution, following the path in `docs/guard/mcp-server-contributions.md`. Discussed with @kantorcodes1 on r/mcp; this is the first memory server in the catalog.

tether is a local-first memory MCP server (SQLite + FTS5, optional libSQL sync). Its state-changing surface is `remember` (upsert), `link`, `forget` (soft delete via `valid_to`, reversible), and `recall`, which updates association weights as a side effect of reading.

## What changed

- `contributions/mcp-servers/mcp.tether.json`: new entry. `uvx` + package `tether-memory`; all four tools plus `other` are `inherit`, so enabling the item keeps Guard's usual review rather than hard-blocking anything. Risk classes: `external-filesystem-write` (the store lives outside the workspace) and `remote-state-mutation` (only when the operator opts into libSQL sync).
- `contracts/extensions/trust-class-map.v1.json`: `command.mcp-tether` added to `external`.
- `pyproject.toml` force-include and `scripts/release/stage_guard_cloud_review_artifacts.py`: package the new JSON.
- `docs/guard/extensions/README.md`: regenerated with `scripts/render_command_extension_directory.py`.
- `tests/fixtures/extension-controls/catalog-baseline.v1.json`: regenerated identity snapshot (catalog digest, 67 → 68 extensions, 222 → 223 permissions).

## Tests

- `tests/test_guard_mcp_server_contribution.py`: tether catalog item is external / opt-in / off by default, launches via `uvx tether-memory`, and every tool resolves to `inherit`.
- `tests/test_guard_mcp_server_grants.py`: tether stays inert until a local-admin enable; a signed-cloud enable does not activate it; a `pipx run tether-memory` launch matches the same package identity; a this-device custom grant still wins over the contribution; and enabling the item leaves `evaluate_tool_call` unchanged for `forget` (same action, source never `catalog-mcp-extension`).
- `tests/test_guard_extension_trust.py`: expected external id set updated.

Ran `ruff check` / `ruff format --check` on touched files, the focused test files, the full `pytest` suite, and `python -m build --wheel` (the wheel contains `mcp_servers/contributions/mcp.tether.json`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Wyui2ch8btSWpCziWFMSvP

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in Tether Memory integration with tools for remembering, recalling, linking, and forgetting information.
  * Tether Memory actions remain subject to review when enabled and are classified as external operations.
* **Documentation**
  * Added the integration to the specialized tools documentation, including its opt-in status and protection model.
* **Tests**
  * Added coverage for trust classification, enablement behavior, tool permissions, launch configuration, and custom permission overrides.

### DIFF
--- a/contracts/extensions/trust-class-map.v1.json
+++ b/contracts/extensions/trust-class-map.v1.json
@@ -78,6 +78,7 @@
     "external": [
       "command.blitcp",
       "command.mcp-filesystem",
+      "command.mcp-tether",
       "command.noodle",
       "command.probe",
       "command.remote.essh",
--- a/contributions/mcp-servers/mcp.tether.json
+++ b/contributions/mcp-servers/mcp.tether.json
@@ -1,0 +1,42 @@
+{
+  "schemaVersion": "guard.mcp-server-contribution.v1",
+  "id": "mcp.tether",
+  "version": "1.0.0",
+  "name": "tether memory MCP",
+  "description": "Reviews tether memory writes: remember, link, forget, and the association updates recall makes. Off until you turn it on.",
+  "trustClass": "external",
+  "activation": "opt-in",
+  "publisher": {
+    "id": "community.sidyellur",
+    "displayName": "sidyellur",
+    "url": "https://github.com/sidyellur/tether"
+  },
+  "icon": {
+    "kind": "react-icon",
+    "name": "HiMiniCube",
+    "background": "#0F766E"
+  },
+  "homepage": "https://github.com/sidyellur/tether",
+  "license": "MIT",
+  "launch": {
+    "kind": "package-launcher",
+    "command": "uvx",
+    "package": "tether-memory"
+  },
+  "riskClasses": ["external-filesystem-write", "remote-state-mutation"],
+  "tools": [
+    { "name": "remember", "state": "inherit" },
+    { "name": "recall", "state": "inherit" },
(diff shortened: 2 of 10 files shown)


kantorcodes (dismissed):
One focused test gap remains before I am comfortable approving the MCP contribution: the current state test can pass when a required Tether tool is missing because `mcp_tool_state()` returns `inherit` for an absent tool. Please first assert that `remember`, `recall`, `link`, and `forget` are actually declared, then assert their inherited states; keep the unknown-tool fallback as a separate assertion.

After that, rerun the MCP contribution/registry/catalog checks. This is a small fix, but it closes a test hole that could otherwise let the published tool boundary silently shrink.

kantorcodes (changes requested):
The test can pass even if a required tool is missing because `mcp_tool_state()` returns `inherit` for absent tools. Assert `remember`, `recall`, `link`, and `forget` are declared first, then check their states. Keep the unknown-tool fallback separate.

PR #2853, opened 2026-09-08
fix: wire Everyday explanations into live review surfaces

kantorcodes:
## Summary

- attach Core-generated `guard.action-explanation.v1` payloads when approval requests are serialized
- render the typed explanation in the live review card instead of leading with raw command/path content when the contract is available
- preserve the existing review presentation as the compatibility fallback when no typed envelope exists
- add regression coverage that verifies the serialized Everyday explanation is identity-bound, path-safe, and does not expose exact command text by default

## Why

The Everyday Mode projector and Dashboard explanation component were present on `main`, but neither had a production caller. This wires the existing Core-owned semantic contract into the actual approval payload and review surface without changing policy, enforcement, approval eligibility, receipts, retention, or entitlements.

## Validation

CI plus the focused Everyday projection/serialization tests must pass before merge.

## Summary by CodeRabbit

- **New Features**
  - Approval cards now display a concise explanation of the requested action when available, including everyday and technical details.
  - Explanations can include relevant action identity, headline, and safely presented path information.

- **Bug Fixes**
  - Approval requests now consistently provide explanation data to the dashboard, including an explicit empty state when no explanation is available.
  - Existing action details and working-directory information remain visible when no explanation is provided.


kantorcodes:
Superseded by #2855. Retargeting this branch exposed unrelated main-to-release history, so I rebuilt the Everyday live wiring cleanly from release/3.2 and carried the review fixes there.

PR #2888, opened 2026-09-11
Add sandbin command safety extension

ayazdoruck:
## Summary

Adds a command safety Extension for [sandbin](https://github.com/ayazdoruck/sandbin), a CLI that runs untrusted code through a local or remote sandbox. Prompted by a discussion about which sandbin CLI commands are safe for an unattended agent to run unreviewed versus which ones need a stop first.

Modeled on `command_repo2nb_extensions.py` per the extension contribution guide:

- **`command.sandbin.run-server`** reviews `sandbin run --server`/`-s` submissions — these POST a new job to a remote sandbin instance and consume its queue/rate-limit quota. `--reconnect <runId>` is an explicit safe variant: it re-attaches to a run the server already accepted instead of calling `POST /runs` again, so it creates no new job.
- **`command.sandbin.keys-create`** reviews `sandbin keys create`, which always mints a new API key against a remote server.
- Left deliberately unmatched by any rule (quiet by design, not by suppression): `sandbin run` with no `--server` (executes locally, never touches the network), `sandbin keys status <key>` (read-only quota check), and `sandbin languages` (reads local host capability, no network call at all).

## Scope

This is intentionally a **focused, minimal first pass** — a draft, not a claim of full coverage against the contribution guide's complete required test matrix (wrappers, quoting, reordered flags, malformed input, etc.). Happy to expand it if a maintainer wants that before merge; wanted to get the core classification boundary in front of you first since it's what came up in discussion.

## Changes

- New `src/codex_plugin_scanner/guard/runtime/command_sandbin_extensions.py`, registered in `command_builtin_extension_catalog.py`.
- Added `command.sandbin` to `contracts/extensions/trust-class-map.v1.json` under `trusted-library` (matching other single-purpose third-party CLIs like `restic`/`rclone`, rather than `external`, since there's no separate `contributions/extensions/*.json` manifest here).
- Regenerated `docs/guard/extensions/README.md` and `catalog.v1.json` via `scripts/render_command_extension_directory.py` / `scripts/export_extension_directory.py`.
- Regenerated the extension-control catalog baseline fixture (`tests/fixtures/extension-controls/catalog-baseline.v1.json`) and updated the one hardcoded total-rule-count assertion in `tests/test_guard_command_extensions.py` (232 → 234).
- One focused classification test: `tests/test_guard_command_sandbin_extensions.py`.

## Test plan

- [x] `uv run pytest -q tests/test_guard_command_sandbin_extensions.py`
- [x] `uv run pytest -q tests/test_guard_command_extension_registry.py tests/test_guard_command_extension_directory.py`
- [x] `uv run python scripts/render_command_extension_directory.py --check`
- [x] `uv run ruff check` / `ruff format --check` on all touched Python files
- [x] Broader slice: `uv run pytest -q -k "catalog or registry or extension_directory or builtin or trust"` — 909 passed. (The only failures in a full run are 2 pre-existing `test_frozen_codex_runtime.py` cases that fail identically on a clean `main` checkout with no changes applied — a local file-permission artifact, unrelated to this change.)

## Summary by CodeRabbit

* **New Features**
  * Added optional safety reviews for sandbin commands that submit remote sandbox jobs or create API keys.
  * Added risk details, examples, and safer alternatives for covered commands.
  * Local runs, read-only commands, and supported reconnect operations remain unflagged; unsupported reconnect syntax is reviewed.

* **Documentation**
  * Added sandbin to the extension directory with its External, opt-in protection model and coverage details.


ayazdoruck:
Good catch — fixed in ffa7bf37b. Moved `command.sandbin` from `trusted-library` to `external`, added the missing `contributions/extensions/command.sandbin.json` manifest (matching `command.repo2nb.json`'s shape), regenerated the directory/catalog/baseline artifacts, and rewrote the classification test to assert against `BUILT_IN_COMMAND_EXTENSION_REGISTRY.observations()` directly plus an explicit inert-until-enabled check — the same pattern `command_repo2nb`'s own test already uses for an opt-in extension, since `evaluate_command()`/`inspect_command()` don't enforce it by default anymore. Re-ran the full trust/contribution/registry/directory slice (936 passed).

ayazdoruck:
Both gitar-bot and Qodo caught the same real gap, thanks — fixed in 43e1dabe0. `safe_flag_variant()`'s generic option parser normalizes `--reconnect=<id>` to the same effective flag as a space-separated `--reconnect <id>`, but sandbin's own argv parser (a plain switch on exact tokens) never recognizes the `=`-joined spelling at all — it falls through to sandbin's unknown-option branch and aborts before any network call. Crediting that spelling as a proven-safe reconnect was unearned.

Replaced it with a custom `_SandbinBareReconnectToken` matcher (same shape as `command_repo2nb`'s own `Repo2nbUnresolvedExpansionMatcher` for an analogous edge case) that only credits a standalone `--reconnect` token, combined via `AllMatcher` with the base `--server` matcher. A `--server` command carrying `--reconnect=X` now stays a genuine review case instead of being silently exempted.

Added `test_sandbin_reconnect_equals_spelling_does_not_earn_the_safe_variant` (fails against the old code, passes against this fix) plus a `--reconnect=<id>` case in the review table. Regenerated the catalog export and baseline fixture — only `catalog_digest` moved. Re-ran the full trust/contribution/registry/directory/sandbin slice: 938 passed.

ayazdoruck:
Good catch, both of you — real regression from my previous fix. Fixed in 887f4dd66.

`_SandbinBareReconnectToken` was doing a naive substring scan for the literal token `--reconnect` anywhere in the argument list, with no awareness of argv position or option-value consumption. `sandbin run --server evil.com -e --reconnect` is a fresh submission whose `--eval` payload literally is the string `--reconnect` — `-e` consumes the next token as its own argument, so `--reconnect` is never a flag there — but the naive scan still found the substring and wrongly exempted it.

Replaced it with `argument_semantics()`, the same positional, consumption-aware parser the base matcher's own `required_flags` check runs on: a value swallowed by a preceding option-with-value never becomes its own effective option. `option_token()` also preserves the exact raw spelling, so the earlier `--reconnect=<id>` fix stays intact in the same check. Verified directly:

```
-e --reconnect        -> option_token(--reconnect) = None
--reconnect abc       -> option_token(--reconnect) = '--reconnect'
--reconnect=abc       -> option_token(--reconnect) = '--reconnect=abc'
```

Added a regression test for this exact case plus the equivalent in the review table. Re-ran the full slice: 939 passed.

kantorcodes (dismissed):
The core reconnect safety bug is fixed on this head, but two merge blockers remain. First, `contributions/extensions/command.sandbin.json` is still not wired into both installed distribution paths (wheel force-includes and frozen release staging), so packaged Guard loses the declared publisher/icon metadata and diverges from source catalog behavior. Second, strengthen the safe-command tests to assert there is no `command.sandbin` registry observation, not only that the inactive default evaluator returns safe; otherwise a safe form can still be silently matched by the opt-in Extension.

Please fix those two active threads, rebase onto current `main`, regenerate the catalog/baselines from the resolved tree, and rerun the focused extension/registry/package tests. After that this should be close.

kantorcodes (changes requested):
`command.sandbin.json` still needs to be included in both installed packaging paths, including frozen staging. Also make the safe-command tests assert there is no `command.sandbin` registry observation, not just a final safe decision.

PR #2915, opened 2026-09-13
fix(daemon): accept the PyInstaller child PID after Desktop spawn

zerocodefast:
## Summary
- Frozen macOS one-file Core writes the extracted app PID into `daemon-state.json`, while Desktop `desktop bootstrap` waits on the bootloader handle from `Popen`.
- Treat that parent/child pair as the same launch so listen-ready daemons can attach instead of timing out before the state file is accepted.

## Testing
- `PYTHONPATH=src python -m pytest tests/test_guard_frozen_daemon_runtime.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon identity verification for frozen, single-file deployments by recognizing trusted child processes launched by the expected daemon.
  * Prevented authentication when the process relationship cannot be verified or the parent process is unrelated.

* **Tests**
  * Added coverage for valid and invalid parent-child process relationships, missing process information, and mismatched daemon identities.


kantorcodes (approved):
The frozen spawn PID match is the missing attach path for macOS one-file Core. Tests cover parent/child matching and POSIX parent parsing. Kilo is not a merge blocker.

PR #2921, opened 2026-09-14
feat(guard): add agi-memory command safety extension

kdbhalala:
## Summary

Adds a command safety extension for [agi-memory](https://github.com/kdbhalala/agi-memory), an MCP memory server for coding agents. It keeps decisions, session history and a code graph in a local SQLite index over an append-only JSONL vault, synced between machines with git.

Opened at the invitation of a HOL Guard maintainer. The v1 boundary settled in review is six reviewed operations — `sync dedupe`, `sync init`, `delete --hard`, `pin`, `unpin`, `bootstrap` — with `sync status`, `log`, `inspect`, `blocks` and `outcome` left automatic, and no prefix-matching on bare `sync`. That is what this implements. I maintain agi-memory, so the CLI behaviour below is verified against the source rather than inferred from `--help`.

## Capability boundary

Extension id `command.agi-memory`, external, opt-in. Six rules, all `default_mode="review"`:

| Rule | Severity | Why it mutates persistent memory |
| :--- | :--- | :--- |
| `command.agi-memory.sync-dedupe` | critical | Rewrites the canonical append-only vault **in place** — the only operation in the tool that rewrites rather than appends. A similarity key that merges two distinct memories destroys the loser, with no undo. |
| `command.agi-memory.delete-hard` | critical | Permanently removes an observation, cascades into the long-term facts promoted from it, and appends a vault tombstone, so the deletion propagates to every other machine on the next sync. |
| `command.agi-memory.sync-init` | high | Attaches a remote to the vault; everything already stored is pushed there on the next sync, and memories routinely quote source and internal decisions. `--create-private` creates a GitHub repository via the `gh` CLI. |
| `command.agi-memory.bootstrap` | medium | Parses git history and README.md and bulk-writes into the store. Against a non-empty store, or the wrong repo, it mixes generated memories into curated ones with nothing marking which is which. |
| `command.agi-memory.pin` | medium | Writes a core-memory block injected into every later session, overwriting whatever that key already held. |
| `command.agi-memory.unpin` | medium | Removes one, dropping an invariant later sessions relied on, with no trace it was ever there. |

Deliberately **not** reviewed: `recall`, `log`, `inspect`, `blocks`, `timeline`, `outcome`, the code-graph queries (`structure`, `callers`, `dependencies`, `impact`) and `sync status`. Side-effect free or purely additive — `outcome` only records how a session ended — and gating them would put a prompt in front of every lookup.

## Two CLI properties that shaped the matchers

Both verified against agi-memory 0.5.0 (`mcp_server.py` dispatch, `sync.py` argparse, `pyproject.toml` console scripts):

**1. Six console scripts, and `bootstrap` has its own pair.** `agent-memory` is an exact alias of `agi-memory`, not a legacy spelling, and both ship in every install — a rule naming only one is bypassed verbatim by typing the other. Separately, the seeder ships as `agi-bootstrap` and `agent-bootstrap`, so a rule covering only the `agi-memory bootstrap` subcommand would miss it completely. All of them are matched, along with `python -m agi_memory.mcp_server`, `python -m agi_memory.bootstrap`, and the usual `exec` / `xargs` wrappers.

**2. `sync` dispatches on a literal positional subcommand.** Only `status|now|dedupe|init` reach the sync module; every other form, **including bare `agi-memory sync`**, falls through to a different program. So `dedupe` is matched as an exact positional, never by prefix.

Two smaller details:

- `--hard` is matched down to `--ha`, the shortest unambiguous argparse prefix. `--h` is deliberately excluded: it is ambiguous with argparse's own `--help`, so it exits without deleting anything, and treating it as destructive would be a false positive.
- Unresolved shell expansions after `sync` or `delete` are reviewed, because they cannot prove the destructive argument absent. Note this overlay covers a **positional** (`dedupe`) as well as a flag (`--hard`).

## Tests

`tests/test_guard_command_agi_memory_extensions.py`, mirroring the repo2nb suite: destructive cases per rule, the `agent-memory` alias for every rule, the standalone `agi-bootstrap` / `agent-bootstrap` scripts, module and wrapper invocations, argparse prefix forms, all four expansion shapes (`$VAR`, `"$VAR"`, `${VAR}`, `` `...` ``, `$(...)`), reordered flags, compound commands where a safe read precedes the destructive segment, action-class ownership, inertness until enabled, and 29 safe reads/help forms asserted non-reviewable, including `outcome`.

## Second commit: catalog artifact resync

Adding an extension moves the catalog identity, so the generated and pinned artifacts move with it. Every value is regenerated by this repository's own scripts and helpers, not hand-edited: `catalog.v1.json`, the extension directory README, the control baseline fixture (extension_count 69 → 70, rule_count 232 → 238, permission_count 245 → 251 — one permission derived per rule), the signed-cloud projection digest vector, and the two release digests pinned in `test_policy_bundle_delivery_runtime.py`.

**One item there is a real fix and deserves a maintainer's eye.** `_rule_version_registry()` in `tests/test_guard_extension_control_authority.py` took `extensions[0].rules[0]` alongside `extensions[0].permissions[0]` and assumed the two lined up. They line up only by coincidence: `permissions_for_rules()` sorts derived permissions by permission id, while rules keep declaration order. The assumption held until an extension whose id sorts first also declared its rules in a different order from their permission ids — which `command.agi-memory` does. The helper now versions the rule that actually owns the permission under test, which is what the test means. Happy to split this out if you would rather take it separately.

## Validation

```
uv run pytest -q tests/test_guard_command_agi_memory_extensions.py              7 passed
uv run pytest -q tests/test_guard_command_extension_registry.py \
               tests/test_guard_command_extension_directory.py                 20 passed
uv run pytest -q -k "extension or catalog or directory or trust or contribution"
                                                          2166 passed, 4 failed
uv run python scripts/render_command_extension_directory.py --check            clean
uv run python scripts/export_extension_directory.py                            committed
uv run python -m ruff check <changed files>                                    clean
uv run python -m ruff format --check <changed files>                           clean
```

The 4 failures in that subset are **identical on unmodified `main`** and untouched by this branch: `test_guard_custom_extension_suggestion` (×2), `test_guard_github_actions_read_workflow`, `test_guard_local_cli_trust`. I verified that by running the same selection on a clean checkout of `main` before and after.

## Notes for review

- Publisher is listed as `community.agi-memory` / "Community", mirroring how repo2nb is listed. Happy to switch to a GitHub-attributed profile per `publisher-metadata.md` if you prefer.
- Icon is `HiMiniCube` from the existing allowlist; no allowlist change requested.
- Argument-level MCP gating (`memory_sync(action=dedupe)` vs `status`, `memory_record` only with `supersedes`) is out of scope here, as discussed. On the agi-memory side the MCP server now ships `readOnlyHint`/`destructiveHint` annotations on `tools/list`, so a future MCP-side gate can read the coarse classification off the server instead of hardcoding tool names.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in external protection extension for `agi-memory` and `agent-memory` commands.
  - Detects six potentially destructive operations, including hard deletion, synchronization changes, bootstrapping, and pin/unpin actions.
  - Recognizes direct commands, Python module launches, shell wrappers, compound commands, and expandable arguments.
  - Provides severity details, safer alternatives, and safe help-command variants.

- **Documentation**
  - Documented the new extension and its six protection rules in the guard extensions guide.


kantorcodes (dismissed):
This PR is not ready yet. The active findings include multiple real bypasses: unresolved expansions are not covered consistently for Windows launcher names or for top-level mutating subcommands (`pin`/`unpin`/`bootstrap`), and valid `exec`/`xargs` value-taking options can shift the wrapped command so protected operations evade review. There are also correctness gaps around help handling (`--help` with expansion evidence and missing `-h`).

The contribution/runtime metadata is also incomplete: packaged builds lose the manifest overlay, the contribution omits protected bootstrap executables/actions, and the runtime spec leaves executable/ecosystem identity empty.

Please fix the matcher/wrapper/expansion gaps first, make manifest/runtime/package metadata describe the same six-rule boundary, regenerate the catalog/projection artifacts, and rerun focused Extension/registry/trust/package/type checks. This needs another full security pass after those changes.

kantorcodes (changes requested):
Expansion handling still misses Windows launchers and top-level `pin` / `unpin` / `bootstrap`, and wrapper value options can shift the protected command. Fix those plus `-h`/help handling. The manifest, runtime spec, and installed package metadata also need to describe the same six-rule surface.

PR #2934, opened 2026-09-15
feat(desktop-core): publish Linux x64 sidecar updates

zerocodefast:
## Summary
- Publish digest-verified Linux x64 Desktop Core sidecars from attested manylinux wheels.
- Let frozen Linux Core installs apply those sidecars with SHA-256 checks instead of Apple identity.

## Testing
- `uv run --frozen pytest tests/test_update_desktop_core.py tests/test_desktop_core_alpha_feed_security.py tests/test_verify_desktop_core_attestation.py -q`
- `uv run --frozen ruff check` and `uv run --frozen ruff format --check` on the changed Python files

## Summary by CodeRabbit

* **New Features**
  * Added support for Desktop Core updates on glibc-based Linux x86-64 systems.
  * Added an immutable Linux x86-64 Desktop Core sidecar and stable release feed.
  * Linux sidecars use digest-based verification and omit Apple signing metadata.
  * Release asset checks now support both macOS ARM and Linux x86-64 builds.

* **Bug Fixes**
  * Unsupported Linux environments are no longer treated as eligible for Desktop Core updates.
  * Linux x86-64 installations no longer use the alpha channel.


kantorcodes (approved):
Approved. Linux Desktop Core sidecar feed is the right product: Desktop should track published CLI sidecars instead of locking Linux to the packaged copy. Remaining review threads are resolved, required checks are green, and the Windows native-wheel flake is unrelated.

PR #2935, opened 2026-09-15
feature Guard: add aether-vault commands to the savety extention

leon1706-lol:
Guard-side extension only (no Aether-Vault API changes):
- Reviews av commit / av push / av gc; keeps status / diff / list-meta quiet and other potentialy dangerous commands 
-  Mirrors command_repo2nb_extensions.py. Includes one focused test.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in safety checks for Aether-Vault commands that publish, promote, or delete model and dataset state.
  - Added warnings for seven higher-risk actions, with guidance for inspection and safer alternatives such as `--dry-run`.
  - Read-only Aether-Vault commands remain unflagged.
  - Added Aether-Vault to the external trust classification and extension catalog.

- **Documentation**
  - Documented the new Aether-Vault extension and its protection rules.

- **Tests & Validation**
  - Added coverage for command detection, safe variants, wrappers, module invocations, trust classification, and generated catalog validation.


kantorcodes (dismissed):
Blocking before approval on the current head:

1. The launcher matrix is incomplete for valid wrapped Python-module forms. `exec python -m av_cli.main push` / `xargs python -m av_cli.main push` (and the supported python/python3/py + `av_cli.main`/`av_cli.launcher` combinations) bypass the protected rules. These are reachable launch forms, so this is a coverage gap, not just test polish.
2. The `exec` option model does not treat `-a` as consuming a value. `exec -a av push` can therefore be misclassified as an Aether-Vault `av push` invocation even though `push` is the executed program. Add wrapper-specific value-option handling and a negative regression.
3. Fix the `_help_variant` return type to `CommandSafeVariant` so the contribution satisfies the repo's type gate, and strengthen the focused tests to assert the final decision-plane action is `review` for every protected case.

The manifest/trust/catalog/package shape is otherwise on the right track. Please address the active review threads, rerun the focused Extension/registry/trust/type checks, and request re-review on the new head.

kantorcodes (changes requested):
`exec python -m ...` / `xargs python -m ...` can bypass the protected rules, and `exec -a` is parsed incorrectly. Also fix `_help_variant` to return `CommandSafeVariant` and make the protected-case tests assert the final `review` action. Please add regressions for the wrapper cases.

PR #2942, opened 2026-09-16
fix(release): treat missing registry projects as empty version lists

zerocodefast:
## Summary
- Treat a missing PyPI or TestPyPI project JSON document as an empty version list in `list-versions`.
- Keep fail-closed behavior for invalid payloads and non-404 errors so main can still compute the next stable version from PyPI and git tags when TestPyPI has no `hol-guard` project.

## Test plan
- [x] Registry listing tests: missing project returns `[]` without retries; invalid data and network errors still fail closed
- [ ] After merge, confirm `Publish to PyPI` on `main` creates the next GitHub and PyPI release

## Summary by CodeRabbit

* **New Features**
  * Added a command for listing available release versions for a registry project.
  * Supports project-specific version listings, canonical version formatting, and sorted results.

* **Bug Fixes**
  * Projects that are not found now return an empty version list instead of an error.
  * Temporary registry access failures are retried automatically, while permanent errors remain clearly reported.
  * Invalid registry responses now produce clear verification errors.


kantorcodes (dismissed):
Approved. Missing TestPyPI project JSON should be an empty version list so main can keep publishing from PyPI and git tags.

kantorcodes (dismissed):
Approved. Missing TestPyPI project JSON is an empty version list, and listing is split out of the oversized verify modules.

kantorcodes (approved):
Approved. Script-mode list-versions now imports listing lazily, so the verifier no longer circular-imports.

PR #2955, opened 2026-09-17
fix(desktop): show policy status without changing local state

kantorcodes:
Shows current connection and policy status without starting background services or creating a new session.

Focused Desktop and policy-status tests pass, and deterministic decision verification is current.


kantorcodes:
Review record retained during sensitive-history cleanup. The earlier automated review marked both findings resolved: read-only status checks must not rewrite receipt rollups, and the Desktop projection must read the verified bundle’s rollout state.

The rollout-state requirement uses Core’s existing policy_bundle_rollout_state helper for v1 and v2, followed by safe-string validation before emitting policyRolloutState. Its regression covers the v2 state under payload.spec. This replacement preserves those findings and their recorded states; final-head CI remains a separate merge requirement.

PR #2986, opened 2026-09-18
New feature

Krishmal2004:
Adds an opt-in Guard command extension, `command.tui-runner`, that reviews TUI Runner's `--reconfigure` flag — the one argv-level flag that forces the setup wizard to run and unconditionally overwrites a project's existing `tui.config.json` once the wizard completes.

TUI Runner's other mutating actions (killing processes bound to configured ports, scaffolding a new project, spawning configured dev commands) are driven entirely through in-process keyboard menu selections after launch. Guard's command hook only sees the argv an agent adapter passes to `hol-guard hook` before execution, so those interactive, post-launch actions are not observable command-line events and are explicitly out of scope for this extension — the extension description, README entry, and a dedicated test call this out rather than claiming coverage that doesn't exist.

Modeled directly on `command_remote_essh_extensions.py` for shape and conventions.

## What's covered

- `tui-runner --reconfigure` (and `.exe`/`.cmd` variants, relative/absolute paths) → flagged as `tui-runner forced reconfiguration command`, severity `medium`, risk class `destructive_shell`.
- Safe/no-op cases confirmed to stay silent: bare `tui-runner`, `--help`, near-miss flags (`--reconfigur`, `reconfigure` without `--`), and mentions inside `echo`/`grep` wrapper commands.

## What's explicitly not covered (documented, not silently dropped)

- Port cleanup, process management, and project scaffolding — all interactive-only, no observable argv.

## Changes

- `contributions/extensions/command.tui-runner.json` — new contribution manifest (opt-in, `external` trust class, `destructive_shell` risk class, safer-alternative guidance).
- `src/codex_plugin_scanner/guard/runtime/command_tui_runner_extensions.py` — new detector module with the `CommandSafetyRule` for `--reconfigure` and the extension spec.
- `src/codex_plugin_scanner/guard/runtime/command_builtin_extension_catalog.py` — registers the new extension in the built-in catalog.
- `src/codex_plugin_scanner/guard/runtime/command_action_risk_classes.py` — maps the new action class to `destructive_shell`.
- `src/codex_plugin_scanner/guard/extension_builder/listing.py` — categorizes `command.tui-runner` under `specialized-tools`.
- `contracts/extensions/trust-class-map.v1.json`, `docs/guard/extensions/catalog.v1.json`, `docs/guard/extensions/README.md`, `tests/fixtures/extension-controls/catalog-baseline.v1.json` — catalog/docs/fixture regeneration for the new extension.
- `tests/test_guard_command_tui_runner_extensions.py` — new tests: local-admin enable/disable gating, safe-command non-triggering, risk-class mapping, and an assertion that the extension's own description documents the interactive-coverage limitation.
- `tests/test_guard_extension_trust.py` — trust-class coverage for the new extension.

## Test plan

- [ ] `pytest tests/test_guard_command_tui_runner_extensions.py -v`
- [ ] `pytest tests/test_guard_extension_trust.py -v`
- [ ] Full test suite to confirm no catalog/fixture regressions

## Summary by CodeRabbit

* **New Features**
  * Added opt-in protection for TUI Runner `--reconfigure` commands.
  * Detects direct, wrapped, reordered, shell-expansion, and Windows executable variants.
  * Classifies forced reconfiguration as a destructive shell action.
  * Provides guidance to review saved configuration before it is overwritten.
  * Added TUI Runner to the specialized tools directory.

* **Documentation**
  * Clarified that interactive menu actions, including port cleanup and project scaffolding, are not covered.


kantorcodes (dismissed):
The new rule still needs the contract cases that prove it cannot be bypassed through wrappers or compound shell forms. Add table-driven coverage for exec/xargs, separators/pipelines, reordered or quoted arguments, and unresolved/malformed input that must stay fail-secure, plus the matching safe counterparts and an evidence/privacy assertion. The current test file only exercises direct argv forms.

Krishmal2004:
@kantorcodes 1. Packaging gap - command.tui-runner.json is now registered in both pyproject.toml's wheel force-include table and scripts/release/stage_guard_cloud_review_artifacts.py's staging map, so installed and frozen builds no longer lose its publisher/icon metadata.

2. Bypass-resistance gap - the underlying matcher actually didn't resist exec/xargs wrappers or unresolved shell expansions (I verified this empirically before writing tests), so I hardened command_tui_runner_extensions.py to match the established repo2nb pattern: it now recognizes exec tui-runner/xargs tui-runner wrapper forms and treats any $VAR/${VAR}/$(...)/backtick expansion in the invocation as fail-secure (reviewed, since it could resolve to --reconfigure). Added table-driven tests covering wrapper bypasses, separators/pipelines, reordered/quoted arguments, unresolved-expansion fail-secure cases, their safe counterparts, and an evidence/privacy assertion proving matcher evidence never leaks raw argument content.

Regenerated the two derived artifacts that depend on the rule description (docs/guard/extensions/catalog.v1.json and the catalog_digest in the fixture tests/fixtures/extension-controls/catalog-baseline.v1.json) so their tests stay in sync. All related test files pass; the only failures I saw (os.geteuid in test_guard_extension_control_proof.py) are pre-existing Windows/Python 3.14 incompatibilities unrelated to this change (confirmed via git stash).

Krishmal2004:
@kantorcodes 
Both files now parse as single valid JSON documents, and the related test suite passes.

Summary of fixes (addressing the CodeRabbit findings):
1. docs/guard/extensions/catalog.v1.json - was two concatenated JSON documents (invalid JSON, Extra data error). Regenerated with python scripts/export_extension_directory.py; now one valid document with 72 entries including mcp.instapods.
2. tests/fixtures/extension-controls/catalog-baseline.v1.json - had a duplicate catalog_digest key, and the underlying baseline was actually stale (mismatched catalog_digest, extension_count: 71 vs actual 72, permission_count: 248 vs actual 249) after the mcp.instapods extension was added. Regenerated the fixture from the live registry; test_guard_extension_control_catalog_detail.py (previously failing on catalog_digest mismatch) now passes.

All related tests (test_guard_extension_control_catalog_detail.py, test_guard_extension_directory_export.py, test_guard_extension_directory_contract_parity.py) pass.

kantorcodes (changes requested):
`TuiRunnerUnresolvedExpansionMatcher` still compares direct launchers to bare `tui-runner` and wrapper targets to a bare prefix, so expansion cases using `tui-runner.exe` or `.cmd` can still miss review. Normalize the executable positions with `executable_names()` and add direct, `exec`, and `xargs` expansion tests for both suffixes.

kantorcodes (changes requested):
Regenerate `contracts/extensions/native-command-program.v1.json` for the new TUI Runner detector. This head changes the command catalog and contribution metadata but leaves the native program artifact untouched, so the packaged Rust path would not include the new rule. Run `uv run python scripts/build_native_command_program.py`, commit the artifact, then make `uv run python scripts/build_native_command_program.py --check` pass.

Krishmal2004:
> Regenerate `contracts/extensions/native-command-program.v1.json` for the new TUI Runner detector. This head changes the command catalog and contribution metadata but leaves the native program artifact untouched, so the packaged Rust path would not include the new rule. Run `uv run python scripts/build_native_command_program.py`, commit the artifact, then make `uv run python scripts/build_native_command_program.py --check` pass.

@kantorcodes sure

Krishmal2004:
@kantorcodes please review and merge

PR #2991, opened 2026-09-18
feat(extensions): add community aparta identity-scope protection

lucascarvalhal:
## Summary
- Add community-maintained `command.aparta` coverage for [aparta](https://github.com/lucascarvalhal/aparta), a CLI that binds one Git, GitHub CLI, gcloud and AWS identity to each project folder and injects it into terminal AI agents. The contribution is external and opt-in, so it stays off until a local-admin enable.
- Review the three command families that leave aparta's read-only surface: `apply`, `remove`, `init`, `add`, `doctor --fix`, `shell-install` and `fallback --secure|--restore` rewrite profile records and identity files on the host (high, `destructive_shell`); `run --profile` / `run -p` executes with another profile's credentials, the one deliberate way to cross the folder's identity scope (critical, `execution`); and `--with-gh-token` on `run` or `env` materializes the GitHub keyring token as `GITHUB_TOKEN` (critical, `local_secret_read`).
- `scan`, `doctor` without `--fix`, `status`, `check`, `list`, `help`, `login`, `env` without the token flag, `run` without `--profile`, the bare `fallback` report, `--dry-run` previews and `--help` stay outside the review boundary and are covered as enabled safe counterparts, including wrappers, reordered flags, quoted operands and launcher suffixes. Since aparta 0.9.3 the inspection commands never write (0.9.2 could reapply a stale profile at startup from any interactive command; that now only happens inside the commands that write anyway). `update` is left to the package extensions: its reapply step runs as `aparta apply --all`, which this rule reviews.
- Publisher metadata is `community.lucascarvalhal`, the aparta author; the listing authorizes GitHub numeric ID `97261310`. CLI behavior was taken from aparta 0.9.x `src/aparta/cli.py`.

## Testing
- `uv run pytest -q tests/test_guard_command_aparta_extensions.py`
- `uv run pytest -q tests/test_guard_command_extension_registry.py tests/test_guard_command_extension_directory.py tests/test_guard_extension_contribution.py tests/test_guard_extension_trust.py tests/test_guard_extension_directory_export.py tests/test_guard_extension_control_catalog_detail.py tests/test_guard_command_extensions.py tests/test_guard_command_catalog_rule_count.py tests/test_policy_bundle_delivery_runtime.py tests/test_guard_command_decision_diff.py`
- `uv run python scripts/export_extension_directory.py --check` and `uv run python scripts/render_command_extension_directory.py --check`
- `uv run python tests/guard_command_decision_diff.py --write`; baseline fixture and projection digest vector regenerated
- `uv run ruff check` and `uv run ruff format --check` on every touched Python file
- Disabled: `hol-guard command explain 'aparta apply client-a' --json` reports no match. Enabled evaluation harness: `aparta apply client-a` -> `command.aparta.identity-write`, `aparta run -p client-a -- ls` -> `command.aparta.scope-crossing`, `aparta env --with-gh-token` -> `command.aparta.token-export`, `aparta doctor` -> no aparta evidence.


kantorcodes (changes requested):
The boundary misses current 0.9.2 identity writes. `aparta init` runs the setup/apply wizard, `doctor --fix` performs repairs, and `add` persists workspace/provider state then reapplies Git, agent, and shell config. Cover those mutation paths with the same fail-secure/safe-preview discipline, or prove why each is outside this extension, and add enabled safe-counterpart tests for the final boundary.

lucascarvalhal:
Good catch, thank you. Pushed a follow-up:

- `init`, `add`, `doctor --fix` (and `-f`) and `shell-install` now join `apply`, `remove` and `fallback --secure|--restore` under `command.aparta.identity-write`, with `--dry-run` and `--help` as safe counterparts, and enabled safe-counterpart tests for each.
- The remaining surface is proven read-only rather than assumed. `doctor` without `--fix`, `scan`, `status`, `check`, `list`, `help`, `env` without `--with-gh-token`, `run` without `--profile`, `login` and the bare `fallback` report never write. That was not quite true in aparta 0.9.2, where any interactive command could reapply a stale profile at startup; aparta 0.9.3 (out shortly) moves that reapply into the commands that already write, so inspection commands only print a notice. The boundary comment in the module documents this, including why `update` is left to the package extensions: its reapply step runs as `aparta apply --all`, which this rule reviews.

PR #2999, opened 2026-09-20
fix(daemon): recover sqlite SELECT failures without killing listen-ready

zerocodefast:
## Summary
- Recover fatal SQLite errors that happen after a connection already opened, including stores whose schema still looks current.
- Keep daemon startup going when AIBOM inventory persist hits a storage error after listen-ready.
- Skip first Cloud sync admission on SQLite failures so dashboard GET `/v1/runtime` does not 500 the Desktop warmup path.

## Test plan
- [x] `pytest` on sqlite failure containment, daemon listen-ready, sqlite start recovery, state persistence, and first-cloud-sync skip cases
- [ ] Confirm a listen-ready daemon stays up if a later SELECT hits a malformed store
- [ ] Confirm Desktop can still open the loopback dashboard after that recovery

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resilience when local storage encounters database errors during synchronization checks.
  - Ensured storage recovery is attempted consistently after serious database failures.
  - Prevented inventory context persistence errors from interrupting daemon operation.
  - Added diagnostics for failed inventory context persistence.
  - Preserved underlying database errors when recovery also fails, improving troubleshooting.

- **Tests**
  - Added coverage for database-failure handling, recovery behavior, and error reporting.


kantorcodes (approved):
Re-approve last push after required checks passed.

PR #3000, opened 2026-09-20
fix(cursor): stop booting frozen core on every hook event

zerocodefast:
## Summary
- Launch managed Cursor hooks through an isolated interpreter so each editor event talks to the resident daemon instead of booting the frozen Guard binary.
- Rebind live frozen `__guard-cursor-hook` commands when that interpreter is available.
- Keep the frozen launcher only as a last resort when no isolated interpreter is present.

## Testing
- `uv run --no-sync pytest -q tests/test_cursor_frozen_hook_command.py tests/test_cursor_hook_rebind.py tests/test_cursor_hooks.py::test_install_cursor_hooks_registers_after_shell_observer`
- `uv run --no-sync ruff check src/codex_plugin_scanner/guard/adapters/cursor_hook_config.py src/codex_plugin_scanner/guard/cursor_hook_rebind.py tests/test_cursor_frozen_hook_command.py tests/test_cursor_hook_rebind.py`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Cursor hook reliability by selecting a validated, isolated Python interpreter when available.
  - Hook commands now run in isolated mode to reduce interference from local Python environment settings.
  - Automatically rebinds hooks that use outdated or frozen control-plane commands.
  - Preserves support for explicitly configured interpreters and supported launcher fallbacks.
  - Ensures fallback and recovery processes are terminated after timeouts, including descendant processes.
  - Uses safer default decisions when policy services are unavailable, while preserving read-event and recording-only access.


kantorcodes (dismissed):

kantorcodes (approved):
Re-approve last push after required checks passed.

